PR #15, opened 2024-12-14
Fix/add weact h562 core board dir to continue board bringup

filipembedded:
Add necessary board directory with **weact_stm32h562_core** board-related files. Necessary for continuing work on board bringup


DNedic:
We want to add support for this board to the `zephyr` repo and upstream it eventually as well, we don't need it in `efc`.

PR #17, opened 2024-12-15
Add Mavlink common and minimal headers and xml files

filipembedded:
Added pregenerated headers in gen directory inside src/protocol/mavlink/ ,with following xml files for common and minimal

### DIFF
--- a/src/protocol/mavlink/common.xml
+++ b/src/protocol/mavlink/common.xml
@@ -1,0 +1,7272 @@
+<?xml version="1.0"?>
+<mavlink>
+  <include>minimal.xml</include>
+  <version>3</version>
+  <dialect>0</dialect>
+  <enums>
+    <enum name="FIRMWARE_VERSION_TYPE">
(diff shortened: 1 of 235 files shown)


mileticveljko (commented):
This pre-generated MAVLink headers are needed for initial development of Sitl!

DNedic:
I am of the opinion that we shouldnt generate anything inside the project source dir and that every build artefact (which generated sources are) should go to `build`. 

Additionally, the generation should be a part of the build process, and not require manual intervention, as this could cause a discrepancy between the source (message sources) and the build result, causing possible confusion and missed bugs when commiting changes. (this especially comes into play with the CI builds)

Please take a look at how this is done in the [ULOG support branch](https://github.com/eurus-project/efc/tree/feat/add_ulog_support).

Finally, a small nitpick, `src/messages/` is to me a bit nicer than `src/protocol`, but this is a personal taste thing.

mileticveljko:
The idea is for pre-generated headers to be removed when that generation is moved into build process. This generated code will indeed be generated into something like `build` directory that is ignored by git.

mileticveljko:
This pre-generated mavlink headers in the source code was originally my idea - this will be done another way.

PR #18, opened 2024-12-15
feat(logging): Add initial ULOG logging support

### LINKED ISSUE #14: Add initial data logging support
In order to analyze sensor/output/system data post-flight or detect erroneous behavior, we need to add logging support. Several options were considered for the logging format, notably rolling a custom format, Tlog and ULOG and ULOG was chosen as the preffered one due to good support for parsing and log analysis software, its self-describing feature and compactness.

DNedic:
This adds initial support for the ULOG logging format. We will use this format to store flight data, as it is self-describing, widely supported for parsing, plotting and analysis software, feature complete, has easy recovery in case of power cuts and has generally low data overhead during the data phase.

Because of a need to create format strings for all messages in order to make the log self-descriptive, code generation was chosen as a preffered option to avoid this error-prone and tedious process. This was done using python and has been integrated into the CMake build process.

`.yaml` was chosen as the format for describing messages, in order to avoid creating custom formats, and because it is easy to read and write for humans.

Initial support does not contain support for:
1. Nested messages
2. Appending data to existing logs
3. Padding - this isn't really relevant for writers and is mostly a reader concern. As most software will manage to read messages without specifically aligned fields just fine, we will only revisit this if necessary.

Closes #14


tools/ulog_gen.py
```diff
@@ -0,0 +1,284 @@
...
+    Message_Header_Type header = {
+        .msg_type = 'F',
+        .msg_size = strlen(format_string),
+    };
+
+    int ret = fs_write(&log->file, &header, sizeof(header));
+    if (ret < 0) {
+        return ULOG_FILESYSTEM_ERROR;
+    }
+
```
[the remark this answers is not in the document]
DNedic: I definitely agree, if we ever have different logging formats we would need to re-do all locking, and doing it at a higher level is more transparent and easier to change. The next step is probably to have a separate logging thread and send data to it using queues.

lkijanovic (approved):
It's overall quite a good initial implementation. I left few comments regarding some future considerations.

tools/ulog_gen.py
```diff
@@ -0,0 +1,284 @@
...
+    words = value.split('_')
+    return '_'.join(word.capitalize() for word in words)
+
+def check_field_types(message):
+    valid_types = ["int8_t", "uint8_t", "int16_t", "uint16_t", "int32_t", "uint32_t",
+                   "int64_t", "uint64_t", "float", "double", "char", "bool"]
+
+    for field in message["fields"]:
+        if field["type"] not in valid_types:
+            raise SystemExit(f"Field {field["name"]} in {message["name"]} message is of unsupported type - {field["type"]}")
```
nikolaptr, line 217: Error is generated while building the project using west
Screenshot from 2024-12-16 17-07-04

It appears that the issue is because of usage of nested double quotation marks in f-string
Issue can be fixed by replacing outer double quotation marks with '

nikolaptr (changes requested):

tools/ulog_gen.py
DNedic: Interesting, I haven't encountered this error at all, but you are right, thanks very much for finding this!

DNedic:
@lkijanovic brought up a valid point on storing timestamps. Ideally, we want timestamps from the moment the data was acquired, however in the current implementation, timestamps of the moment the data is logged are used. These two could differ based on a variety of factors, for instance if a separate logging thread is used, and it is of lower priority than say control or estimation threads, logging the data could lag behind data acquisition.

DNedic:
The approach in regards to timestamping was changed. Instead of storing the log-time timestamp, we are now always adding a timestamp field to the message strucuture and requiring users to fill it in with the acquisition-time timestamp.

src/logging/ulog.c
```diff
@@ -0,0 +1,333 @@
...
+
+    return ULOG_SUCCESS;
+}
+
+ULOG_Error_Type ULOG_Sync(ULOG_Inst_Type *log) {
+    if (log == NULL) {
+        return ULOG_INVALID_PARAM;
+    }
+
+    int ret = WriteSync(&log->file);
```
nikolaptr, line 304: Phase guard should be added for ULOG_Sync message, function shouldn't allow user to write Sync out of Data Phase of the ULog

nikolaptr (changes requested):

src/logging/ulog.c
DNedic: Good find, thanks!

nikolaptr (changes requested):

src/logging/ulog.c
```diff
@@ -0,0 +1,364 @@
...
+        return ULOG_WRONG_PHASE;
+    }
+
+    int ret = WriteSync(&log->file);
+    if (ret < 0) {
+        return ULOG_FILESYSTEM_ERROR;
+    }
+
+    ret = fs_sync(&log->file);
+    return ret == 0 ? ULOG_FILESYSTEM_ERROR : ULOG_SUCCESS;
```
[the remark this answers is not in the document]
DNedic: Thanks!

DNedic:
Thank you for reviews @lkijanovic @nikolaptr , I will add unit tests and API documentation in a follow-up PR.

nikolaptr (approved):
LGTM

PR #19, opened 2024-12-22
add GH-actions CI stage for building project

mileticveljko:
The CI pipeline now includes a build stage, implemented using GitHub Actions.

This stage first installs Zephyr and other required dependencies, followed by the build process for the `blackpill_f411ce` device. The build stage is triggered on every push to the remote repository and on every merge to the master branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,67 @@
+name: build
+
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install required packages
+        run: |
+          sudo apt install --no-install-recommends -y git cmake ninja-build gperf \
+            ccache dfu-util device-tree-compiler wget \
+            python3-dev python3-pip python3-setuptools python3-tk python3-wheel xz-utils file \
+            make gcc gcc-multilib g++-multilib libsdl2-dev libmagic1 python3-venv
+
+      - name: Create a virtual environment for Zephyr and intall west
+        run: |   
+          python3 -m venv $GITHUB_WORKSPACE/zephyrproject/.venv
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          pip install west
+
+      - name: Get Zephyr source code
+        run: |
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          west init -m https://github.com/eurus-project/zephyr $GITHUB_WORKSPACE/zephyrproject/
+
+      - name: Update Zephyr Workspace
+        run: |
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          cd $GITHUB_WORKSPACE/zephyrproject/
+          west update
+          
+      - name: Export Zephyr CMake Package
+        run: |
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          cd $GITHUB_WORKSPACE/zephyrproject/
+          west zephyr-export
+
+      - name: Install Python dependencies
+        run: |
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          cd $GITHUB_WORKSPACE/zephyrproject/
+          west packages pip --install
+
+      - name: Install Zephyr SDK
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+            ccache dfu-util device-tree-compiler wget \
+            python3-dev python3-pip python3-setuptools python3-tk python3-wheel xz-utils file \
+            make gcc gcc-multilib g++-multilib libsdl2-dev libmagic1 python3-venv
+
+      - name: Install Zephyr
+        run: |
+          python3 -m venv $GITHUB_WORKSPACE/zephyrproject/.venv
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          pip install west
+          west init -m https://github.com/eurus-project/zephyr $GITHUB_WORKSPACE/zephyrproject/ > /dev/null 2>&1
```
DNedic, line 22: What is the reason behind redirecting to `stderr`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: zephyrproject/${{ github.repository }}
+      
+      - name: Build project
+        run: |
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
+          cd $GITHUB_WORKSPACE/zephyrproject/${{ github.repository }}
+          west update
```
DNedic, line 43: We don't have a `west.yml` file nor any external dependencies yet, perhaps it is better to defer adding this step for later.

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+          python3 -m venv python_venv
+          source python_venv/bin/activate
+          pip install -r requirements.txt
+          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
```
DNedic, line 47: We don't actually need to activate the environment (and it might also conflict with zephyrs venv as well). Things which require our venv will call its python directly. We might want to reevaluate this approach in the future, but for now it's good enough.

DNedic:
Thank you for adding this @mileticveljko, LGTM overall!

.github/workflows/build.yml
mileticveljko: I redirected both `stdout` and `stderr` of two commands with excessively large outputs, which made the stage logs difficult to read. These commands are part of the Zephyr installation process, and if they fail, the failure will still be reported.

To investigate errors of those commands, you can, for example, create a separate branch with commit that removes `> /dev/null 2>&1`. These investigation branches are not intended for merging into the main project and can be deleted after troubleshooting.

DNedic: I don't think we should do this, large outputs are not a problem as on GitHub you have sorted output step-wise. Also, needing to create separate branches to diagnose this adds a lot of friction to debugging when something fails.

mileticveljko: I honestly find 10.000+ repeated lines of `Updating repository(x/y)` really annoying and scrolling through them while working on this was painful.
I assume that, It is better to put those commands in separate steps - As you said. GitHub will sort them.

DNedic: That is a good solution IMO. Personally wheneve I've had to deal with CI logs, searching for `error` would have been enough, so the number of lines never bothered me.

mileticveljko: I am going to continue working on this - to provide better organization of steps

I am marking this PR as **draft**, it shouldn't be merged until this process is completed

mileticveljko:
This PR is no longer a **draft**

DNedic:
LGTM, please squash the commits and switch to [conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/) and we can merge this.

mileticveljko:
Done

DNedic:
> Done

I'm sorry to be overly pedantic, but a proper tag in this case would be `ci` or `change`, feat is reserved for user-facing features.

mileticveljko:
No problem
It's not bothering to change commit message

DNedic (approved):
LGTM

PR #21, opened 2024-12-29
change: Add formatting config and pre-commit hook

DNedic:
This adds the proposed formatting rules and a corresponding `pre-commit` hook, as well as a `Contributing` section of the README.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,5 @@
+repos:
+- repo: https://github.com/pre-commit/mirrors-clang-format
+  rev: v18.1.8
+  hooks:
+  - id: clang-format
--- a/README.md
+++ b/README.md
@@ -18,3 +18,13 @@ pip install -r requirements.txt
 ```bash
 west build -b blackpill_f411ce
 ```
+
+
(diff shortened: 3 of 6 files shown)


mileticveljko:
If I understand correctly, this applies to all files that are in the commit phase.
It might be handy for this PR to change all files by running `pre-commit run --all-files', to make sure the format is applied to all files from this point on.

Also, any change to the `.clang_format` file should be followed by a `pre-commit run --all-files' to apply the change immediately to every file - Maybe it is handy to point that in README.

All in all, this is a great thing to have!

DNedic:
>It might be handy for this PR to change all files by running `pre-commit run --all-files`, to make sure the format is applied to all files from this point on.

True, I have assumed that the files are formatted *enough* considering a formatting pass previously for `main.c` and that I just reformated `ulog.h`, leaving only `ulog.c`, I will still do `pre-commit run --all-files` just in case.

> Also, any change to the `.clang_format` file should be followed by a `pre-commit run --all-files` to apply the change immediately to every file - Maybe it is handy to point that in `README`.

This is a very nice suggestion, but the problem is that reformatting all files would ruin git blame for all of them. This can be solved manually, by adding a [.git-blame-ignore-revs](https://github.com/espressif/esp-serial-flasher/blob/master/.git-blame-ignore-revs) file, so we should keep this a manual process IMO in order not to ruin git blame accidentally.

Thanks for the review!

mileticveljko (approved):
Great!

Issue #11, opened 2024-12-13
Add CI checks


DNedic:
In order to avoid introducing regressions, either functional, style or policy we should consider adding a CI pipeline.

Some things we could add to the CI:
1. Make sure our firmware builds
2. Make sure we don't introduce any sources with incompatible licenses or forget to add license notices
3. Enforce formatting rules
4. Possibly run static analysis and/or other tools to find issues

mileticveljko:
Do you think GitHub Actions meets our requirements?

Perhaps this Issue can include adding Cmake target for code formatting and possibly for adding license notice to source files without one - so that process is automated for developers.

DNedic:
Usually this process is automated outside of the build system, using tools like [pre-commit](https://pre-commit.com/) (see example [here](https://github.com/espressif/esp-serial-flasher/blob/master/.pre-commit-config.yaml)) and I do think we should add this as well.

DNedic:
Code formatting and `pre-commit` hook is added [here](https://github.com/eurus-project/efc/pull/21), if/when merged, we should add a GitHub workflow to run `pre-commit` hooks in the CI.

mileticveljko:
GitHub action that run `pre-commit` in CI added. PR: [#22](https://github.com/eurus-project/efc/pull/22)

PR #22, opened 2024-12-29
ci: add format check stage to ci pipeline

mileticveljko:
The CI pipeline now includes a format check stage, implemented using GitHub Actions.

This stage uses clang-format (_v18.1.3_) to check that code is formatted regularly. The format is specified in the `.clang-format` file.

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -1,0 +1,16 @@
+name: clang-format check
+
+on: 
+  push:
+  
+jobs:
+  formatting-check:
+    name: format check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Run clang-format style check for C/C++ source files.
+      uses: jidicula/clang-format-action@v4.14.0
+      with:
+        clang-format-version: '18'
+        check-path: 'src'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v18.1.8
+  rev: v18.1.3
   hooks:
   - id: clang-format


DNedic:
We shouldn't do this separately, as each time we have to match versions, and there will also be more `pre-commit` checks in the future, we can simply use pre-commit itself in the CI like so: `pre-commit run --show-diff-on-failure --from-ref master --to-ref HEAD` and save ourselves the work from having to separately maintain these between the `pre-commit` config and the CI.

mileticveljko:
CI format check now uses pre-commit.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 repos:
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v18.1.8
+  rev: v18.1.3
```
DNedic, line 3: Why change the version to go back to a previous bugfix release?

.github/workflows/format-check.yml
DNedic, line 1: We should rename this file to `pre-commit` or `pre-commit-checks`, as there will be other pre-commit checks in the future.

.github/workflows/format-check.yml
```diff
@@ -0,0 +1,32 @@
...
+  push:
+  
+jobs:
+  formatting-check:
+    name: format check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install requirements
+        run: |
+          sudo apt install --no-install-recommends -y git python3-venv clang-format
```
DNedic, line 13: We don't actually need `clang-format` here, `pre-commit` will pull it from a git tag.

.github/workflows/format-check.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Create a virtual environment and intall pre-commit
+        run: |   
+          python3 -m venv $GITHUB_WORKSPACE/.venv
```
DNedic, line 17: Do we need a venv here? A docker container is created solely for this check, and is killed after, we only need pre-commit here and have no other python packages to have dependency issues with.

.github/workflows/format-check.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          path: ${{ github.repository }}
+
+      - name: Run pre-commit
+        run: |
+          cd $GITHUB_WORKSPACE/${{ github.repository }}
+          $GITHUB_WORKSPACE/.venv/bin/pre-commit run -a --show-diff-on-failure
```
DNedic, line 28: While yes, if we do a manual pass for all pre-commit checks once, in the future everything done by pre-commit is the difference between the commit in question and anything before, this still can pose problems for some other checks.

It's better to do `--from-ref HEAD` + `--to-ref master` or such.

.github/workflows/format-check.yml
mileticveljko: Thank you - didn't know that

.github/workflows/format-check.yml
mileticveljko: You're right. We, actually do not need virtual env

.github/workflows/format-check.yml
mileticveljko: Shouldn't it be the other way around: `--from-ref master` `--to-ref HEAD`

.github/workflows/pre-commit-checks.yml
```diff
@@ -0,0 +1,32 @@
+name: clang-format check
```
DNedic, line 1: Please change the workflow name to match the filename as well.

.github/workflows/format-check.yml
DNedic: Looks like you're correct, thanks!

DNedic (approved):

mileticveljko:
#11

PR #23, opened 2024-12-29
feat(baro): Add support for BMP280 Barometer

nikolaptr:
Add BMP280 sensor data acquisition, altitude calculation and appropriate barometer ULog message.

### DIFF
--- a/messages/ulog/baro.yaml
+++ b/messages/ulog/baro.yaml
@@ -1,0 +1,9 @@
+name: baro
+description: Contains barometer temperature and pressure.
+fields:
+  - name: temperature
+    type: float
+    description: Temperature [degC]
+  - name: pressure
+    type: float
+    description: Pressure [kPa]
--- a/src/main.c
+++ b/src/main.c
@@ -15,6 +15,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <math.h>
 #include <stdio.h>
 #include <zephyr/device.h>
 #include <zephyr/drivers/gpio.h>
@@ -28,7 +29,15 @@
 #include <zephyr/usb/usbd.h>
 
 #include "ulog.h"
+#include "ulog_altitude.h"
+#include "ulog_baro.h"
 #include "ulog_gyro.h"
+
+#define ALTITUDE_SOURCE_TYPE_BARO 1
+
+// Value derrived from BMP280 datasheet (page 15 of 49) for the given sensor
+// configuration.
+#define BMP280_ALTITUDE_VARIANCE_M .017f
 
 LOG_MODULE_REGISTER(main);
 
@@ -50,6 +59,8 @@ static struct fs_mount_t main_fs_mount = {
 
 static ULOG_Inst_Type ulog_log;
 static uint16_t gyro_msg_id = 0;
+static uint16_t baro_msg_id = 0;
+static uint16_t baro_alt_msg_id = 0;
 
 #if defined(CONFIG_USB_DEVICE_STACK_NEXT)
 static struct usbd_context *sample_usbd;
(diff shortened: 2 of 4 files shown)


messages/ulog/baro.yaml
```diff
@@ -0,0 +1,9 @@
+name: baro
+description: Contains barometer temperature, pressure and altitude
+fields:
+  - name: temp
+    type: float
+  - name: press
+    type: float
+  - name: alt
```
DNedic, line 8: Other values are raw sensor readings, but this is an estimated (and possibly filtered) altitude and should not be in the same message. In the future the altitude will be estimated from multiple sources as well.

messages/ulog/baro.yaml
```diff
@@ -0,0 +1,9 @@
+name: baro
+description: Contains barometer temperature, pressure and altitude
+fields:
+  - name: temp
+    type: float
+  - name: press
```
DNedic, line 6: Please either add units in the comment, or in the variable name (`press_bar`).

src/main.c
```diff
@@ -118,6 +121,54 @@ static int process_imu(const struct device *dev) {
     return 0;
 }
 
+static int process_baro(const struct device *dev) {
+    struct sensor_value temp;
+    struct sensor_value press;
+
+    float temp_f;
```
DNedic, line 128: What does the `_f` suffix stand for?

DNedic:
Nice work @nikolaptr , I will test this with real hardware soon.

messages/ulog/baro.yaml
DNedic: Additionally, there is no reason to abbreviate these IMO. `pressure` and `temperature` are short enough.

messages/ulog/baro.yaml
nikolaptr: I even considered completely omitting altitude from the log, as it is currently redundant (directly derived from pressure), but kept it for the initial phase

We could create `AltEst` ULog message with `src` (estimation source), `est` (estimation) and `var` (variance) fields. Where altitude estimation source could be defined using `ULOG_AddInfo()`

messages/ulog/baro.yaml
nikolaptr: Thanks!

src/main.c
nikolaptr: Stands for the float value, used to distinguish from the `temp` and `press`

DNedic: With modern code editors or IDEs, finding a type of a variable is a simple hover over it, I don't think we need this personally.

messages/ulog/baro.yaml
DNedic: That would be good, it can be just called 'Altitude' since altitude is always estimated anyway (and again, no need to abbreviate that much).

src/main.c
nikolaptr: I agree, thanks!

DNedic:
Everything seems to work fine on real hardware.

src/main.c
```diff
@@ -28,8 +29,16 @@
...
 #include "ulog.h"
+#include "ulog_altitude.h"
+#include "ulog_baro.h"
 #include "ulog_gyro.h"
 
+#define ALTITUDE_SOURCE_TYPE_BARO 1
+
+// Value derrived from BMP280 datasheet (page 15 of 49) for the given sensor
+// configuration.
+#define BMP280_ALTITUDE_VARIANCE_M .017f
```
DNedic, line 40: Generally, we wouldn't want sensor or chip-specific things at this high of a level, but it will suffice for now.

DNedic: Perhaps we should think about creating maps between sensor ids and their properties (such as variance) soon.

DNedic:
LGTM

DNedic (approved):

PR #25, opened 2025-01-05
added venv_python path for windows platform

kikolino28:
Python virtual environments on the Windows platform have a different structure than on other platforms. The bin folder does not exist, and python.exe is located in the Scripts folder. Because of this, the path to Python in src/logging/ulog/CMakeLists.txt must be specified differently if CMake detects the Windows platform.


DNedic:
Thanks for fixing the windows building issue, please add a PR description describing the logic behind the changes and motivation in the future.

DNedic:
Also please use the agreed upon [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/), we don't have a linter to check these yet but we will add it in the future.

DNedic:
Thanks a lot for fixing this!

DNedic (approved):

PR #27, opened 2025-01-12
Added accel and mag ulog messages

filipembedded:


src/main.c
```diff
@@ -211,6 +221,10 @@ int main(void) {
         LOG_ERR("Could not register ULOG gyro format!");
     }
 
+    if (ULOG_Accel_RegisterFormat(&ulog_log) != ULOG_SUCCESS) {
+        LOG_ERR("Could not register ULOG gyro format!");
```
DNedic, line 225: Accel format

openocd_stm32f4_stlink.cfg
DNedic, line 1: Why is this change bundled with accel changes?

messages/ulog/gyro.yaml
DNedic, line 1: Let's try to leave newlines at the end of textual files.

DNedic:
LGTM overall, nice work!

DNedic:
Please use the conventional commit format. Also we should strive to have PR and commit descriptions to explain why something is done.

openocd_stm32f4_stlink.cfg
filipembedded: I shouldn't have done this here. There is some issue with transport layer name in .cfg file. I will make separate PR for that.

messages/ulog/gyro.yaml
filipembedded: Is there some specific reason for that?

DNedic: A quick google gave me [this](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) but I remember programs giving me warnings about them for the longest time.

messages/ulog/accel.yaml
```diff
@@ -0,0 +1,9 @@
+name: accel
+description: Contains accelerometer data in yaw, pitch and roll format
+fields:
+  - name: x
+    type: float
+  - name: y
+    type: float
+  - name: z
+    type: float
```
nikolaptr, line 9: Please add units for the each field of the message, either in the commend or the variable name (i.e. x_g).
This should also be done for the rest of messages

filipembedded: > Please add units for the each field of the message, either in the commend or the variable name (i.e. x_g). This should also be done for the rest of messages

I  will add unit  for each field in comments, I think that it looks much cleaner that way

src/main.c
```diff
@@ -113,7 +115,15 @@ static int process_imu(const struct device *dev) {
         .z = sensor_value_to_float(&gyro[2]),
     };
 
+    ULOG_Accel_Type accel_msg = {
+        .timestamp = k_uptime_get() * 1000,
```
nikolaptr, line 119: Data from gyroscope and accelerometer need to have the same timestamp value, because they are data from the same IMU sample.

messages/ulog/mag.yaml
```diff
@@ -0,0 +1,14 @@
+name: mag
+description: Contains magnetometer data in yaw, pitch and roll format
+fields:
+  - name: x
+    type: float
+    description: Magnetic induction on x axis [G]
+
```
nikolaptr, line 7: Newlines are not needed here IMO

nikolaptr (commented):
LGTM, great job!
Please just squash these commits into one to keep things tidy

openocd_stm32h5_stlink.cfg
```diff
@@ -1,4 +1,4 @@
-source [find interface/stlink-v2.cfg]
+source [find interface/stlink.cfg]
```
DNedic, line 1: Can we remove this from this PR?

src/main.c
```diff
@@ -345,12 +260,10 @@ int main(void) {
...
-        process_baro(main_baro);
-
         ULOG_Sync(&ulog_log);
 
         k_msleep(1000);
     }
 
     return 0;
-}
+}
```
DNedic, line 269: Please terminate the file with a newline.

DNedic:
Please rebase on top of `master`, you're overwriting changes merged in the meantime.

openocd_stm32h5_stlink.cfg
filipembedded: Yes

messages/ulog/mag.yaml
filipembedded: It looked better to me, but if it is the issue, I will change it

filipembedded:
This has been solved in another PR, no need to have this no more

PR #32, opened 2025-01-19
Replace deprecated stlink_v2.cfg file with relevant stlink.cfg in openocd_stm32h5_stlink.cfg file

filipembedded:
The use of stlink_v2.cfg is deprecated, use stlink.cfg in openocd_stm32h5_stlink.cfg

### DIFF
--- a/openocd_stm32h5_stlink.cfg
+++ b/openocd_stm32h5_stlink.cfg
@@ -1,4 +1,4 @@
-source [find interface/stlink-v2.cfg]
+source [find interface/stlink.cfg]
 
 transport select hla_swd 
 


DNedic:
I think we should actually completely remove these openocd config files, they are not tied to the project itself and people can use custom configs as they wish. We can make a folder for them and add it to .gitignore. In addition to that we should investigate using west as a wrapper for starting debug sessions as boards in Zephyr usually already contain parts of openocd configuration related to them.

filipembedded:
> I think we should actually completely remove these openocd config files, they are not tied to the project itself and people can use custom configs as they wish. We can make a folder for them and add it to .gitignore. In addition to that we should investigate using west as a wrapper for starting debug sessions as boards in Zephyr usually already contain parts of openocd configuration related to them.

I agree that openocd.cfg files are not related to project itself, and using west wrapper to debug is definitely a neater way. Shall we open an issue about this?

DNedic:
I've created https://github.com/eurus-project/efc/issues/33 to track that. You can feel free to create a new PR to remove all of these and create a `.gitignore` rule when you have the time.

PR #34, opened 2025-01-20
feat(SBUS): Add support for SBUS

nikolaptr:
Added support for receiving aircraft setpoint data via SBUS.


filipembedded (approved):
Looks good to me, good job!

src/main.c
```diff
@@ -183,6 +192,50 @@ static int process_baro(const struct device *dev) {
...
+
+    if (!device_is_ready(sbus_dev)) {
+        LOG_ERR("Device %s not found. Aborting!.", sbus_dev->name);
+        return;
+    }
+
+    while (1) {
+        k_sem_take(&sbus_sync, K_FOREVER);
+
+        printf("Received SBUS data:\n"
```
filipembedded, line 231: Is this data converted to percentage of throttle, or it is raw data?

src/main.c
```diff
@@ -183,6 +192,50 @@ static int process_baro(const struct device *dev) {
     return 0;
 }
 
+static void sbus_event_callback(struct input_event *evt, void *user_data) {
```
DNedic, line 195: In which context is this callback called? Interrupt, another thread or something else?

src/main.c
```diff
@@ -62,6 +63,14 @@ static uint16_t gyro_msg_id = 0;
 static uint16_t baro_msg_id = 0;
 static uint16_t baro_alt_msg_id = 0;
 
+static const struct device *const sbus_dev =
+    DEVICE_DT_GET(DT_CHOSEN(futaba_sbus));
+static struct k_sem sbus_sync;
+static int32_t sbus_throttle;
```
DNedic, line 69: It's not enough to have a semaphore to protect these globals AFAIK. If you want to access them from two contexts you either need a mutex or to make them atomic.

I would suggest not using a semaphore here but instead a queue or some other method of IPC.

src/main.c
DNedic: Good point, if we are already printing these, they should be in human-readable form, not raw values.

src/main.c
```diff
@@ -183,6 +192,50 @@ static int process_baro(const struct device *dev) {
...
+    }
+
+    if (evt->sync) {
+        k_sem_give(&sbus_sync);
+    }
+}
+
+INPUT_CALLBACK_DEFINE(sbus_dev, sbus_event_callback, NULL);
+
+void sbus_thread(void *, void *, void *) {
```
DNedic, line 220: Let's split every important thread into its own sourcefile in `src`. This as a general rule will make our code much more navigable.

src/main.c
nikolaptr: Thanks for pointing this out, not all receivers output data in the full scale range (0-2047) of the SBUS channel.
So converting these channel data to percentages i.e., without calibration procedure, won't do us any good. Calibration procedure should be used first to determine min and max values of each channel, but that cannot be done without having receiver first.

src/main.c
nikolaptr: It is called from another thread, which is dedicated to processing incoming SBUS data

src/main.c
nikolaptr: Thank you, i agree

src/main.c
nikolaptr: I agree, will do!

src/main.c
```diff
@@ -337,6 +341,12 @@ int main(void) {
         .filename = filename,
     };
 
+    k_thread_create(&radio_thread, radio_thread_stack,
+                    K_THREAD_STACK_SIZEOF(radio_thread_stack), radio_receiver,
+                    NULL, NULL, NULL, K_LOWEST_APPLICATION_THREAD_PRIO, 0,
```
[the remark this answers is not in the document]
nikolaptr: I agree, but i left it for now at lowest priority untill we define priorities of other threads. Other option would be to increase it to highest priority for now

src/radio_receiver.c
```diff
@@ -0,0 +1,88 @@
...
+        break;
+    case INPUT_ABS_RZ:
+        radio_receiver_data.yaw_rate = evt->value;
+        break;
+    default:
+        break;
+    }
+
+    if (evt->sync) {
+        k_queue_append(&sbus_queue, &radio_receiver_data);
```
DNedic, line 58: Does this block?

DNedic: Also, is the data copied to the queue or is only the reference added?

src/radio_receiver.c
```diff
@@ -0,0 +1,88 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <zephyr/input/input.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+#include "radio_receiver.h"
+
+typedef struct radio_receiver_data_t {
```
DNedic, line 24: Why both typedef and declare a new struct? I guess we can just go along with Zephyrs approach and declare structs and enums without typedefing them.

src/main.c
```diff
@@ -76,6 +77,9 @@ static uint16_t accel_msg_id = 0;
 static uint16_t baro_msg_id = 0;
 static uint16_t baro_alt_msg_id = 0;
 
+K_THREAD_STACK_DEFINE(radio_thread_stack, 512);
```
DNedic, line 80: Have you tested the stack usage? 512 is awfully low this early in the development phase and considering how much free RAM we have.

We can always do analysis down the line and reduce all stack sizes.

DNedic:
Terrific job, I guess we will be merging this after it gets tested with real hardware?

src/radio_receiver.c
nikolaptr: Habit, i'll adjust it to Zephyrs approach

src/main.c
nikolaptr: I have tested it, it's enough as the thread just receives data from queue for now (it does not parse data), in the future this will most definitely need to be increased

DNedic: Let's bump it to 1024 then so it doesn't bite us when we forget about it.

src/radio_receiver.c
nikolaptr: It does not, but i am just now reading about Zephyr [Message Queues](https://docs.zephyrproject.org/latest/kernel/services/data_passing/message_queues.html) which appear to be better for our use case, as data is fixed in size

src/main.c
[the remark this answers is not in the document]
DNedic: I guess highest priority would be better.

src/radio_receiver.c
DNedic: I'm sorry if this question is uninformed, but how is the user informed of failure in that case? The function doesn't have a return param.

src/main.c
```diff
@@ -337,6 +341,11 @@ int main(void) {
         .filename = filename,
     };
 
+    k_thread_create(&radio_thread, radio_thread_stack,
+                    K_THREAD_STACK_SIZEOF(radio_thread_stack), radio_receiver,
+                    NULL, NULL, NULL, K_HIGHEST_APPLICATION_THREAD_PRIO, 0,
```
DNedic, line 346: I'm sorry, I just remembered, when used like this, this makes the thread [cooperative (non-preemptible)](https://docs.zephyrproject.org/latest/kernel/services/threads/index.html#thread-priorities). I'm not sure we want this, perhaps we should play it safe and choose the highest preemptible thread priority.

src/radio_receiver.c
nikolaptr: Sorry, on a first glance i thought it was non blocking, but it does block, documentation does not specify this
In background `_append` function calls `SYS_PORT_TRACING_OBJ_FUNC_BLOCKING` with `K_FOREVER` parameter, which results in function blocking indefinitely

Even more of a reason to switch to solution like message queues

DNedic (approved):

nikolaptr:
Thank you for the reviews @DNedic @filipembedded!

PR #37, opened 2025-01-22
feat(ulog): Add accelerometer and magnetometer Ulog messages to complete IMU Ulog messages

filipembedded:
"Added Ulog messages for accelerometer and magnetometer data. Updated the main.c file to include testing for these changes. Verified the Ulog logs using PlotJuggler, focusing only on accelerometer data as the hardware for the magnetometer is currently unavailable. The results look good."

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -29,6 +29,7 @@
 #include <zephyr/usb/usbd.h>
 
 #include "ulog.h"
+#include "ulog_accel.h"
 #include "ulog_altitude.h"
 #include "ulog_baro.h"
 #include "ulog_gyro.h"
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -86,6 +88,8 @@ static int process_imu(const struct device *dev) {
     struct sensor_value temperature;
     struct sensor_value accel[3];
     struct sensor_value gyro[3];
+    uint64_t currentTime;
```
DNedic, line 91: Please use `snake_case` for variables for consistency.

DNedic: Also, it would be good to express the units this is in with a suffix (e.g. `uptime_ms`).

DNedic: And finally there is no reason to define this here and then assign a value later, you can just create a const value and immediately assign it a value.

DNedic:
LGTM! I've only left one nitpick.

DNedic (approved):

Issue #36, opened 2025-01-22
Implement board tests


lkijanovic:
Essentially, the idea is to have a set of hardware tests that should be kept simple and try to isolate a module/functionality as much as possible.
Some of the immediate and eventual benefits of this:

- New features can be tested (relatively) isolated before integrating them in the main app
- When the moment comes (and it will come) that you stop trusting HW, there will be a reference test to check whether you're losing your mind
- During the bring-up of future boards, tests will be a useful starting point

I would probably suggest placing these in _**src/tests/hardware/**_, at least in the first iteration.

DNedic:
This would be nice to have. We could also have a physical machine with boards attached to it to run CI tests on real hardware.

With the current architecture of the repo, the repo itself represents a standalone app and anything inside src is a part of the app, so having other apps in there would not make sense IMO. We could create a new repo for test apps, or rearchitect the repo folder structure (but try preserving git blame).

PR #38, opened 2025-01-23
change(tests): setup PoC infrastructure for hardware tests

lkijanovic:
Simple blinky test along with some CMake changes to allow west to build both the test and the main app (but not simultaneously).
Test can be run with:

`west build -b blackpill_f411ce -p -- -DEFC_TEST_NAME=blinky`

### DIFF
--- a/tests/hardware/blinky/CMakeLists.txt
+++ b/tests/hardware/blinky/CMakeLists.txt
@@ -17,28 +17,8 @@
 
 cmake_minimum_required(VERSION 3.20.0)
 
-find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
-project(efc
-    VERSION 0.1
-    LANGUAGES C
+message("Got into blinky")
+target_sources(app
+PRIVATE
+main.c
 )
-
-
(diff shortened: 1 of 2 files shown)


filipembedded:
I have run this test on real hardware, it works!
Test_Board

DNedic:
The test seems fine but I can only reiterate what I said [here](https://github.com/eurus-project/efc/issues/36#issuecomment-2608492486), I don't think this belongs here. 

We can take several routes to remedy that:
1. Create a separate repo and pull `efc` as a submodule
2. Move the current `efc` app-related files into `apps/efc` and then put the tests into `apps/tests/test_xyz`. Also we should investigate creating a multi-app Zephyr workspace with a `west` manifest.
3. Implement test functions in `efc` and use the [shell](https://docs.zephyrproject.org/latest/services/shell/index.html) interface to trigger them
4. Not do this at all, since we're going to make more errors into non-critical errors in `efc` and so the execution of different parts of code can still execute without some hardware or software parts

Aditionally, using a define passed to CMake seems like a hacky way to do this, even if we would stick to a monolithic CMake setup, we should use CMake targets for this.

PR #40, opened 2025-02-01
ci: fix ci build stage by adding apt update & upgade

mileticveljko:
Build stage of GH actions CI pipeline was missing packages updating, which caused stage breakage. This PR also includes removing of unnecessary activation of python virtual environments and fixing ubuntu to 24.04 version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,10 +5,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Install required packages
         run: |
+          sudo apt update && sudo apt upgrade
           sudo apt install --no-install-recommends -y git cmake ninja-build gperf \
             ccache dfu-util device-tree-compiler wget \
             python3-dev python3-pip python3-setuptools python3-tk python3-wheel xz-utils file \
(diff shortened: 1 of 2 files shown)


lkijanovic (approved):

.github/workflows/build.yml
```diff
@@ -5,10 +5,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Install required packages
         run: |
+          sudo apt update && sudo apt upgrade
```
DNedic, line 12: I believe `-y` is required for the upgarde operation. Also, `apt` commands have an [unstable API](https://manpages.debian.org/unstable/apt/apt.8.en.html), `apt-get` should be used in scripts.

.github/workflows/build.yml
```diff
@@ -5,10 +5,11 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
DNedic, line 8: GitHub caches commonly used images and workflows based on them execute very fast, this bit me once when every Debian workflow I tried to run took over an hour for an open source repo, please check if an older version of ubuntu is cached so this doesn't suprise us.

.github/workflows/build.yml
```diff
@@ -17,37 +18,32 @@ jobs:
...
           python3 -m venv $GITHUB_WORKSPACE/zephyrproject/.venv
-          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
-          pip install west
+          $GITHUB_WORKSPACE/zephyrproject/.venv/bin/pip install west
 
       - name: Get Zephyr source code
         run: |
-          source $GITHUB_WORKSPACE/zephyrproject/.venv/bin/activate
-          west init -m https://github.com/eurus-project/zephyr $GITHUB_WORKSPACE/zephyrproject/
+          $GITHUB_WORKSPACE/zephyrproject/.venv/bin/west init -m https://github.com/eurus-project/zephyr \
```
DNedic, line 25: This is less error-prone but I think it's a bit less readable. Still, we can leave it.

DNedic (commented):
LGTM overall, thanks for fixing this!

Issue #16, opened 2024-12-15
Add mavlink headers (if it is protocol of choise) for Sitl


mileticveljko:
Is MAVLink selected as the ground station messaging protocol?

If so, it would be great if the MAVLink headers directory exists somewhere in ***efc*** source. MAVLink headers are pretty much required (at least _common_ ones) for Sitl to communicate with the JMavSim flight simulator.

And if not, MAVLink headers will be in Sitl source.

It is important to avoid duplication.

DNedic:
We have decided that for now `mavlink` will be the choice of protocol for ground communication, yes, you can proceed with adding these, however you can also add that to the `sitl` branch while developing it as well.

mileticveljko:
Any idea where to put the MAVLink headers in the efc source? Something like `src/protocols/mavlink'? I assume that we will, possibly, have more protocols in the project.

DNedic:
I'd personally vote for `src/telemetry/mavlink`, as there will be other protocols unrelated to telemetry and this grouping makes it easier to grasp the functionality of different repo parts.

mileticveljko:
MAVLink headers added by @filipembedded with branch `fead/add_mavlink_headers`.

mileticveljko:
MAVLink message generating is now on branch `feat/mavlink`.

DNedic:
Closed by https://github.com/eurus-project/efc/pull/41

PR #41, opened 2025-02-02
feat(telemetry): Initial mavlink telemetry support

DNedic:
This adds initial telemetry to ground using the Mavlink protocol, over UART. Mavlink is the reference protocol used in the UAV world to communicate current UAV status to the ground and send commands to the UAV and is supported by a variety of software (e.g. QGroundControl, MissionPlanner), hence the choice for a first protocol was easy and `common.xml` as a first dialect was chosen in order to be compatible with aformentioned software.

UART was chosen as most UAV-oriented RF modules will talk over UART and it is easy to get this data to a PC with an USB to serial converter. Perhaps the baudrate can be bumped later.

Due to the STM32F411 chip we are working with now having only a 1-byte UART FIFO, using TX interrupts was not an option as they would be too CPU costly and data alignment issues prevented the use of DMA, since Mavlink is a variable size message protocol, so a thread was created to send the data in a polling manner sent over a pipe. 

On the build system side, generation of dialects was implemented similar to ULOG message generation, where multiple dialect files can be specified.

Finally, since Mavlink has a few quirks, these had to be handled by implementing helper functions and global buffers.


DNedic:
Closes #16

filipembedded (approved):
This looks very good for the initial phase. Good job 👍

src/main.c
```diff
@@ -48,6 +51,13 @@ LOG_MODULE_REGISTER(main);
 static const struct gpio_dt_spec fw_running_led =
     GPIO_DT_SPEC_GET(DT_ALIAS(led0), gpios);
 
+static const uint8_t telemetry_system_id = 0;
+static const uint8_t telemetry_channel_ground = 0;
+
+static const uint8_t telemetry_component_id_base = 0;
+static const uint8_t telemetry_component_id_imu = 1;
+static const uint8_t telemetry_component_id_baro = 2;
+
```
lkijanovic, line 60: We should probably move these to a separate header at one point, possibly as an enumerated type.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "mavlink"]
+	path = tools/mavlink
+	url = git@github.com:mavlink/mavlink.git
```
lkijanovic, line 3: Just a note - when we move to using a west manifest to manage efc and zephyr repos, this could be handled by manifest file as well. I don't really have a strong opinion on this, I'm just listing options. If it turns out that using git submodule is simpler, there might not be any need to change this.

src/main.c
```diff
@@ -48,6 +51,13 @@ LOG_MODULE_REGISTER(main);
 static const struct gpio_dt_spec fw_running_led =
     GPIO_DT_SPEC_GET(DT_ALIAS(led0), gpios);
 
+static const uint8_t telemetry_system_id = 0;
+static const uint8_t telemetry_channel_ground = 0;
+
+static const uint8_t telemetry_component_id_base = 0;
```
nikolaptr, line 57: My understanding here may be wrong, but i think that component id is assigned per device. Assigning component id's in this way would eventually result in clash with some of the standard component id's defined in [MAV_COMPONENT](https://mavlink.io/en/messages/common.html#MAV_COMPONENT) enum. I think [MAV_COMP_ID_AUTOPILOT1](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_AUTOPILOT1) would be valid here.

Also, componend id 0 is not a valid *source* component id for a message, it is reserved for *target_component* for some of the messages (i.e. [MISSION_REQUEST](https://mavlink.io/en/messages/common.html#MISSION_REQUEST)), when broadcast to all of the components is needed.

src/main.c
```diff
@@ -112,25 +132,31 @@ static int process_imu(const struct device *dev) {
...
-           sensor_value_to_double(&accel[0]), sensor_value_to_double(&accel[1]),
-           sensor_value_to_double(&accel[2]), sensor_value_to_double(&gyro[0]),
-           sensor_value_to_double(&gyro[1]), sensor_value_to_double(&gyro[2]));
+    const uint16_t telemetry_msg_len = mavlink_msg_scaled_imu_pack_chan(
+        telemetry_system_id, telemetry_component_id_imu,
+        telemetry_channel_ground, &mavlink_msg_buf, timestamp_ms,
+        sensor_value_to_milli(&accel[0]), sensor_value_to_milli(&accel[1]),
+        sensor_value_to_milli(&accel[2]), sensor_value_to_milli(&gyro[0]),
+        sensor_value_to_milli(&gyro[1]), sensor_value_to_milli(&gyro[2]), 0, 0,
+        0, sensor_value_to_float(&temperature));
```
nikolaptr, line 141: Temperature of SCALED_IMU message is in cdegC units, conversion must be done.

src/main.c
```diff
@@ -170,32 +196,59 @@ static int process_baro(const struct device *dev) {
...
-        .timestamp = k_uptime_get() * 1000,
+        .timestamp = timestamp_ms * 1000,
         .temperature = temperature,
         .pressure = pressure,
     };
 
+    const uint16_t telemetry_msg_len = mavlink_msg_scaled_pressure_pack_chan(
+        telemetry_system_id, telemetry_component_id_baro,
+        telemetry_channel_ground, &mavlink_msg_buf, timestamp_ms,
+        pressure * (1 / 10.0f), 0.0f, (int16_t)temperature, 0);
```
nikolaptr, line 210: Same as with SCALED_IMU message, SCALED_PRESSURE message requires temperature in cdegC units. Also, pressure must be in hPa, to convert pressure from kPa it must be multiplied by 10.

nikolaptr (commented):
Great job for implementing this! I only have few remarks

.gitmodules
DNedic: I agree that that would probably be a prefferable option at that point, we should look into it.

src/main.c
DNedic: Absolutely, thanks for the suggestion!

src/main.c
DNedic: Good find, this was lacking an explicit cast.

src/main.c
DNedic: This does infact return temperature in degrees Celsius, unless I am missing something.

nikolaptr: MAVLink message requires cdegC, which is centi degC, meaning that the value of 1 equals to 0.01 degC

DNedic: I didn't properly read that haha, there goes mavlink with their weird units again.

DNedic:
Thanks for the reviews @nikolaptr @lkijanovic @filipembedded !

src/main.c
```diff
@@ -112,25 +129,31 @@ static int process_imu(const struct device *dev) {
...
-           sensor_value_to_double(&accel[0]), sensor_value_to_double(&accel[1]),
-           sensor_value_to_double(&accel[2]), sensor_value_to_double(&gyro[0]),
-           sensor_value_to_double(&gyro[1]), sensor_value_to_double(&gyro[2]));
+    const uint16_t telemetry_msg_len = mavlink_msg_scaled_imu_pack_chan(
+        telemetry_system_id, telemetry_component_id, telemetry_channel_ground,
+        &mavlink_msg_buf, timestamp_ms, sensor_value_to_milli(&accel[0]),
+        sensor_value_to_milli(&accel[1]), sensor_value_to_milli(&accel[2]),
+        sensor_value_to_milli(&gyro[0]), sensor_value_to_milli(&gyro[1]),
+        sensor_value_to_milli(&gyro[2]), 0, 0, 0,
+        sensor_value_to_milli(&temperature) / 100);
```
nikolaptr, line 138: This should be multiplied by 100

src/main.c
```diff
@@ -170,32 +193,58 @@ static int process_baro(const struct device *dev) {
...
-        .timestamp = k_uptime_get() * 1000,
+        .timestamp = timestamp_ms * 1000,
         .temperature = temperature,
         .pressure = pressure,
     };
 
+    const uint16_t telemetry_msg_len = mavlink_msg_scaled_pressure_pack_chan(
+        telemetry_system_id, telemetry_component_id, telemetry_channel_ground,
+        &mavlink_msg_buf, timestamp_ms, pressure * 10.0f, 0.0f,
+        sensor_value_to_milli(&baro_temp) / 100, 0);
```
nikolaptr, line 207: This should also be multiplied by 100

src/main.c
DNedic: The value obtained is in milli, so this is fine.

PR #45, opened 2025-02-03
feat(config): Switch the FPU and flash prefetcher on

DNedic:
This ensures the FPU is switched on and we are using the flash prefetcher in order to utilise the hardware better and achieve better performance.


lkijanovic (approved):
I'd be curious to see how much of an improvement is actually achieved, but I guess we don't have much performance benchmarking ATM.

Somewhat unrelated note on the matter of FPU: we probably want to have some of the current floating-point calculations have a fixed-point alternative as well, to better support boards without an FPU. This is of course waaaaaaaay down the road.

nikolaptr (approved):
LGTM!

Issue #31, opened 2025-01-18
Split the logic into multiple threads and establish IPC mechanisms


DNedic:
Currently, we are doing everything in a polling manner in the main thread, which is no way scalable, since different functionality will have different priorities (e.g. logging code should be lower priority than motor control or sensor data acquisition).

Additionally, we should investigate different IPC mechanisms and approaches for different scenarios, notably it looks like [ZBus](https://docs.zephyrproject.org/latest/services/zbus/index.html) could be a good fit for what we are doing.

DNedic:
First steps towards this can be seen in https://github.com/eurus-project/efc/pull/41 and https://github.com/eurus-project/efc/pull/34

lkijanovic:
I'd like to give some initial thoughs on the topic. I'm not certain there is a one-size-fits-all solution for the problem - if there is, it would most likely be ZBus as it's (at least I think) the only mechanism than can handle multiple consumers. This is all quite hypothetical as I haven't done extensive investigation of ZBus pros and cons. Zephyr's FIFOs and pipes also seem to fit the bill quite well for some more specific scenarios so they should be kept in mind as well.

Few things that I think are imporant to consider when designing data paths through the system:

- All data should be timestamped as close as possible to where it originated (sensor read, radio receive, etc).
- Every rule has its exceptions, but generally, for any inter-thread SP-SC link, consumer should do a blocking receive, and producer should do a non-blocking send. This scheme minimizes CPU utilization and allows for easy logging in the case of an overflow.
- It should be explored whether Zephyr has possiblity to allow thread to block on multiple kernel objects (something like poll()/select() with POSIX, or queue sets with FreeRTOS). ZBus would be kindof a workaround for the issue, but alternatives should be explored as well.

DNedic:
I second everything @lkijanovic said, these are the conclusions I came to as well over time and also while working on #41 and have been implemented there.

Additionally, I'll add that if you're deciding on whether something should be a thread or not, think about which operations would be more important than others for a UAV (e.g. sending motor control signals is more important than logging data or sending telemetry in order to keep the vehicle airborne) in places where the dataflow diverges (after receiving a value from a sensor for instance).

Issue #29, opened 2025-01-12
Add ULog messages for IMU data


nikolaptr:
We need to define ULog messages which are going to be used to log IMU (Inertial Measurement Unit) data. 

IMU often consists of an accelerometer, gyroscope and a magnetometer, and can be:
1. 3DOF (accelerometer, gyroscope or magnetometer only)
2. 6DOF (any pair of above mentioned sensors)
3. 9DOF (accelerometer, gyroscope and magnetometer)
Please note that most of the IMU sensors also have the temperature sensing capabilities.

An efficient way of logging IMU data needs to be defined, should IMU data be divided in multiple messages in respects to sensor type, or are all sensor values be combined into single IMU message.

DNedic:
We can take several approaches to unify this data, but they all have drawbacks:
1. Create 6dof and 9dof messages which nest gyro, accel, mag messages
This is a good general DRY approach, but it presents problems like:
* Gyro, accel and mag messages need to contain temperature data if they are used by themselves, this would be thus duplicated
* Any software that uses logs and processes them needs to know whether to read gyro, accel and mag separately, using the 6dof or the 9dof message
* If we have multiple sensors for a specific role (especially if say it is a situation of one 9dof and an extra mag) sensor id handling can get quite hazy (multi_id in ULOG, our own IDs etc)

2. Keep using messages separately
This adds storage and speed overhead, as well as making the code a bit more complex.

3. Create entirely new 6dof and 9dof messages
This has all of the drawbacks except the first one of the first approach, but is no longer DRY (if anything in gyro, accel or mag messages changes, these changes have to be replicated manually in 6dof and 9dof messages)

filipembedded:
> We can take several approaches to unify this data, but they all have drawbacks:
> 
>     1. Create 6dof and 9dof messages which nest gyro, accel, mag messages
>        This is a good general DRY approach, but it presents problems like:
> 
> 
>     * Gyro, accel and mag messages need to contain temperature data if they are used by themselves, this would be thus duplicated
> 
>     * Any software that uses logs and processes them needs to know whether to read gyro, accel and mag separately or using the 6dof or the 9dof message
> 
>     * If we have multiple sensors for a specific role (especially if say it is a situation of one 9dof and an extra mag) sensor id handling can get quite hazy (multi_id in ULOG, our own IDs etc)
> 
> 
>     2. Keep using messages separately
>        This adds storage and speed overhead, as well as making the code a bit more complex.
> 
[...]

PR #49, opened 2025-02-08
refactor: Split into threads and establish ZBUS communication

DNedic:
This splits most of the current logic, notably the logger and telemetry packer into separate threads with adequate priorities, and establishes IPC using ZBUS.

ZBUS has been chosen as an IPC mechanism due to its ability to reduce the amount of copying since the published data exists in one place, be more verbose about data dependencies, provide a lot of convenience features in one place, such as notifications, as well as its ability to collect statistics.

Additionally, there are now defined types for data exchange that do not depend on zephyr and can be used for SITL/unit tests down the line.


app/src/logger.c
```diff
@@ -0,0 +1,198 @@
...
+            ULOG_Altitude_Type altitude_msg = {
+                .timestamp = msg.timestamp_us,
+                .source = ALTITUDE_SOURCE_TYPE_BARO,
+                .altitude = altitude,
+                .variance = BMP280_ALTITUDE_VARIANCE_M};
+
+            ULOG_Altitude_Write(&ulog_log, &altitude_msg, baro_alt_msg_id);
+        }
+
+        ULOG_Sync(&ulog_log);
```
DNedic, line 196: I'm not sure if we should keep this. Perhaps we can run it periodically from another place, but I have yet to investigate if that is thread-safe.

app/src/main.c
```diff
@@ -27,61 +27,66 @@
...
 
-K_THREAD_STACK_DEFINE(telemetry_thread_stack, 2048);
-static struct k_thread telemetry_thread;
+K_THREAD_STACK_DEFINE(telemetry_packer_thread_stack, 2048);
+static struct k_thread telemetry_packer_thread;
+
+K_THREAD_STACK_DEFINE(telemetry_sender_thread_stack, 1024);
+static struct k_thread telemetry_sender_thread;
+
+K_THREAD_STACK_DEFINE(logger_thread_stack, 8192);
```
DNedic, line 88: Logging still requires a big stack, probably due to lfs and SDMMC drivers.

lkijanovic: There are also some local variable that are not that small in logger thread func.

app/src/logger.c
lkijanovic: Periodic sync could be also sent via zbus and caught in logger thread, not sure if that's a waste of resources though...

app/src/logger.c
```diff
@@ -0,0 +1,198 @@
...
+
+    if (ULOG_Baro_Subscribe(&ulog_log, 0, &baro_msg_id) != ULOG_SUCCESS) {
+        LOG_ERR("Could not subscribe ULog baro message!");
+    }
+
+    if (ULOG_Altitude_Subscribe(&ulog_log, 0, &baro_alt_msg_id) !=
+        ULOG_SUCCESS) {
+        LOG_ERR("Could not subscribe ULog altitude message!");
+    }
+
```
lkijanovic, line 136: If it isn't too dificult to achieve, maybe pre-while part of this should be split in a separate function (`logger_init` or something of sorts) and called prior to thread creation.

DNedic: I'm not sure what this would achieve, timing wise.

app/src/logger.c
DNedic: I thought about this approach for heartbeat but it seemed wasteful at the time. I'll reconsider and maybe try using a timer.

DNedic:
Closes #31

lkijanovic (approved):

app/src/logger.c
```diff
@@ -0,0 +1,213 @@
...
+            ULOG_Baro_Type baro_msg = {
+                .timestamp = msg.timestamp_us,
+                .temperature = msg.temperature_degc,
+                .pressure = msg.pressure_kpa,
+            };
+
+            ULOG_Baro_Write(&ulog_log, &baro_msg, baro_msg_id);
+
+            const float altitude =
```
nikolaptr, line 198: Maybe it would be better to keep altitude calculation in `main` for now.

Having it in `logger` section is a bit confusing and complicates things if we wish to transmit [MAVLink Altitude Message](https://mavlink.io/en/messages/common.html#ALTITUDE) at some point

app/src/telemetry_packer.c
```diff
@@ -0,0 +1,135 @@
...
+ZBUS_CHAN_DECLARE(baro_chan);
+
+ZBUS_CHAN_DEFINE(heartbeat_chan, bool, NULL, NULL,
+                 ZBUS_OBSERVERS(telemetry_packer_sub), 0);
+
+ZBUS_SUBSCRIBER_DEFINE(telemetry_packer_sub, 16);
+
+const uint8_t telemetry_system_id = 0;
+const uint8_t telemetry_component_id = MAV_COMP_ID_AUTOPILOT1;
+const uint8_t telemetry_channel_ground = 0;
```
nikolaptr, line 38: `MAVLINK_COMM_0` could be used here instead of `0`

app/src/types.h
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <stdint.h>
+
+struct imu_data {
+    uint64_t timestamp_us;
+    float temperature_degc;
+    float accel_mps2[3];
+    float gyro_radps[3];
```
nikolaptr, line 24: Similar to discussion in #29, it would be better to separate this data into separate structures or add magnetometer data and status (which could indicate which fields contain useful data).

nikolaptr (approved):
Great job! I have only left few remarks

app/src/types.h
DNedic: Good suggestion, but these types can be extended at any time, and are not meant to be perfect from the get-go like ulog and mavlink messages which are important for future parse-ability of old logs.

app/src/telemetry_packer.c
DNedic: Thanks!

app/src/logger.c
DNedic: Thanks for the suggestion, I agree completely, however at the moment I'd leave it in, since this is derived data, and I'd like to see a separate estimator thread spit out positional data in the future after aggregating all sensor data. This is dirty, but works until we get there.

app/src/types.h
DNedic: I'll still add magnetometer data just so we don't forget.

DNedic: Actually, it might be better to just rename this to `imu_6dof_data`.

PR #50, opened 2025-02-09
fix(telemetry): Fixed issue of MAVLink messages not being serialized

nikolaptr:
MAVLink requires messages to be serialized before they are transmitted, this intermediate step was added to ensure proper data is sent.


DNedic (approved):
Good catch, this explains why I haven't been able to use QGCs Mavlink Inspector! LGTM!

DNedic:
One minor thing, please try to use the most general commit message group, for instance `fix(telemetry)` instead of fix(mav)`, then you can specify that the fix is regarding mavlink in the commit title further.

lkijanovic (approved):

PR #54, opened 2025-02-12
feat(icm42688p): Added support for ICM42688P

nikolaptr:
Added support for ICM42688P accelerometer & gyroscope sensor and necessary KConfig option which shall allow user to choose primary IMU.

### DIFF
--- a/app/Kconfig
+++ b/app/Kconfig
@@ -7,4 +7,17 @@ config APP_DATA_LOGGING_SYNC_INTERVAL
 		This configures the interval at which the logging system flushes everything to disk
 		and hence determines the maximum duration of data lost on power cuts.
 
+choice APP_PRIMARY_IMU
+	prompt "Primary IMU"
+	help
+		This configures Eurus Flight Controller primary IMU.
+
+config APP_PRIMARY_IMU_MPU6050
+ 	bool "MPU6050"
+
+config APP_PRIMARY_IMU_ICM42688P
+	bool "ICM42688P"
+
+endchoice
+
 source "Kconfig.zephyr"
--- a/app/boards/blackpill_f411ce.overlay
+++ b/app/boards/blackpill_f411ce.overlay
@@ -15,6 +15,8 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <zephyr/dt-bindings/sensor/icm42688.h>
+
 / {
 	chosen {
 		zephyr,console = &cdc_acm_uart;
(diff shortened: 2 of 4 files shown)


app/Kconfig
```diff
@@ -7,4 +7,21 @@ config APP_DATA_LOGGING_SYNC_INTERVAL
 		This configures the interval at which the logging system flushes everything to disk
 		and hence determines the maximum duration of data lost on power cuts.
 
+choice
```
DNedic, line 10: Please add a choice name. I suggest `APP_PRIMARY_IMU`, but anything similar is fine.

app/Kconfig
```diff
@@ -7,4 +7,21 @@ config APP_DATA_LOGGING_SYNC_INTERVAL
...
+	prompt "IMU"
```
DNedic, line 11: Please add a string representing the option name, such as "Primary IMU".

app/Kconfig
```diff
@@ -7,4 +7,21 @@ config APP_DATA_LOGGING_SYNC_INTERVAL
...
+	help
+		This configures Eurus Flight Controller primary IMU.
+
+config APP_DATA_USE_MPU6050
```
DNedic, line 15: Change these to match the name of the choice and add the IMU name after, for suggested example: `APP_PRIMARY_IMU_MPU6050`. This would be in-line with how Kconfig options are usually done.

app/Kconfig
```diff
@@ -7,4 +7,21 @@ config APP_DATA_LOGGING_SYNC_INTERVAL
...
+ 	bool "MPU6050"
+	help
+		Eurus Flight Controller shall use MPU6050 as primary IMU.
+
+config APP_DATA_USE_ICM42688P
+	bool "ICM42688P"
+	help
```
DNedic, line 22: All of these help strings feel a bit redundant and unnecessary IMO, at that point the user already knows this info. If we have some more context we can add it here tho,

app/boards/blackpill_f411ce.overlay
```diff
@@ -49,17 +51,33 @@
...
+        accel-odr = <ICM42688_DT_ACCEL_ODR_50>;
+        gyro-pwr-mode= <ICM42688_DT_GYRO_LN>;
+        gyro-fs = <ICM42688_DT_GYRO_FS_2000>;
+        gyro-odr = <ICM42688_DT_GYRO_ODR_50>;
+	};
 };
 
 &i2c1 {
-	main_imu: mpu6050@68 {
+	imu_mpu6050: mpu6050@68 {
```
DNedic, line 80: Will anything fail if we select the ICM in Kconfig but leave this here?

app/boards/blackpill_f411ce.overlay
```diff
@@ -49,17 +51,33 @@
     status = "okay";
     pinctrl-0 = <&spi2_sck_pb13 &spi2_miso_pb14 &spi2_mosi_pb15>;
     pinctrl-names = "default";
-    cs-gpios = <&gpiob 12 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>;
+    cs-gpios = <&gpiob 12 (GPIO_ACTIVE_LOW | GPIO_PULL_UP)>,
```
DNedic, line 54: It would be nice to think about how to keep these pins available in cases where we aren't using the ICM, but this is not a blocker.

DNedic (changes requested):
LGTM overall, nice work, some suggestions left.

app/src/main.c
```diff
@@ -210,7 +210,16 @@ int main(void) {
         return 0;
     }
 
-    const struct device *const main_imu = DEVICE_DT_GET(DT_NODELABEL(main_imu));
+#if CONFIG_APP_DATA_USE_MPU6050
+    const struct device *const main_imu =
```
DNedic, line 214: It just ocurred to me, Zephyr DTS bindings exist so we don't have to do this, specifically we can use the `chosen` mechanism and only refer to the device through `DT_CHOSEN` without ever specifying the name of the label itself in sources.

app/boards/blackpill_f411ce.overlay
nikolaptr: MCU will still perform MPU6050 initialization during boot but will not communicate with the device afterwards. In case this initialization fails, error message will be sent via terminal, but the rest of the code will not be affected.

app/boards/blackpill_f411ce.overlay
nikolaptr: I agree!

app/src/main.c
nikolaptr: That would require usage of KConfig options in the scope of .overlay file, i may be wrong but i don't think there is a way to do this, at least not a clean one

app/src/logger.c
```diff
@@ -110,8 +110,13 @@ void logger(void *dummy1, void *dummy2, void *dummy3) {
         LOG_ERR("Could not system name info to the log!");
     }
 
+#if CONFIG_APP_DATA_USE_MPU6050
     const char main_imu_name_key[] = "char[7] main_imu_name";
     const char main_imu_name[] = "MPU6050";
+#elif CONFIG_APP_DATA_USE_ICM42688P
+    const char main_imu_name_key[] = "char[9] main_imu_name";
+    const char main_imu_name[] = "ICM42688P";
+#endif
```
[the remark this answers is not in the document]
nikolaptr: I agree, a mechanism for this must be implemented in the future

app/boards/blackpill_f411ce.overlay
DNedic: That's good enough for now, we just need to be sure it won't stop the execution.

To discuss this, https://github.com/eurus-project/efc/pull/54#discussion_r1953451962 and https://github.com/eurus-project/efc/pull/54#discussion_r1953488619, we're going to be creating a new issue.

DNedic (approved):

lkijanovic (approved):
Looks good to me, leaving aside the overlay/config issues we discussed. However, I feel these are way beyond the scope of this PR and anyway they warrant a lot of additional discussion.

Good job!

nikolaptr:
Thank you for the reviews @DNedic @lkijanovic!

Issue #46, opened 2025-02-04
Update years in the licence statements


lkijanovic:
It's needed for someone to go over the existing files in the repo and update years in the licence preambles. I'm creating this issue as I'm not completely certain what year format is expected. I think that there is also some sort of an option of doing this through GH actions automatically, but it needs to be investigated.

PR #55, opened 2025-02-14
Update years in the licence statements

filipembedded:
In order to resolve the issue opened by @lkijanovic : [(https://github.com/eurus-project/efc/issues/46)] This PR presents my proposal for a solution. The idea is to put a date in form (**year** - Present), where  **year** represents the year when file has been created. The word **Present** would serve as a placeholder for a year when some stable version of this project would be published. Please note that this is just my proposal, and that I am glad to hear another ideas.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 #
 # This file is part of the efc project <https://github.com/eurus-project/efc/>.
-# Copyright (c) 2024 The efc developers.
+# Copyright (c) (2024 - Present), The efc developers.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -1,6 +1,6 @@
 #
 # This file is part of the efc project <https://github.com/eurus-project/efc/>.
-# Copyright (c) 2024 The efc developers.
+# Copyright (c) (2024 - Present), The efc developers.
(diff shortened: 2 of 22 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 #
 # This file is part of the efc project <https://github.com/eurus-project/efc/>.
-# Copyright (c) 2024 The efc developers.
```
DNedic, line 3: Why do we need braces and a comma? Also there is no reason to capitalise "present".

DNedic:
Thanks for working on this, it seems like a good solution so we don't have to update the file headers all the time. I'm not sure on the legality of it though, we have to check if it has some licensing implications.

mileticveljko (approved):
Thank you

CMakeLists.txt
filipembedded: I have seen that structure in some device drivers provided by some very popular vendors (like NXP Semiconductors). I think that comma gives additional "focus" to the developers that work on the specific file. Capitalization of the word 'present' was totally up to me at the moment of writing this, I don't have specific reason for that, I will change it.

filipembedded:
> Thanks for working on this, it seems like a good solution so we don't have to update the file headers all the time. I'm not sure on the legality of it though, we have to check if it has some licensing implications.

Can you help me check this licensing implications, I am totally unexperienced with this. Thanks in advance!

DNedic (approved):

Issue #51, opened 2025-02-09
Make sure the autopilot can operate without a logging fs


DNedic:
Currently, if there is no SD card, or if the filesystem mounting fails, the firmware will stop executing. Adopt a new approach where without the logging fs, the firmware keeps running and use of it will not be attempted.

DNedic:
Closed by #56

PR #56, opened 2025-02-14
feat: Tolerate FS abscence or failure

DNedic:
This makes sure that if fs mounting or file creation fail, we don't abort the whole program.

Additionally, if obtaining boot count fails, logging will proceed to a temporary file.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -63,7 +63,7 @@ struct fs_mount_t main_fs_mount = {
     .mnt_point = "/SD:",
 };
 
-uint32_t boot_count = 0;
+int32_t boot_count = -1;
 
 struct k_pipe telemetry_ground_pipe;
 static uint8_t telemetry_ground_pipe_data[1024];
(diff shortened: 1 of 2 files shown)


DNedic:
Closes #51

app/src/main.c
```diff
@@ -236,48 +279,13 @@ int main(void) {
...
-    }
-
-    ret = fs_read(&boot_count_file, &boot_count, sizeof(boot_count));
-    if (ret < 0) {
-        LOG_ERR("Could not read from file!\n");
-        return 0;
     }
 
-    ret = fs_seek(&boot_count_file, 0, FS_SEEK_SET);
+    ret = update_boot_count();
```
nikolaptr, line 284: What would happen if `update_boot_count` fails but `ULOG_Init` passes?

Maybe it would be safer to move `update_boot_count` to `logger.c` and perform a check there?

nikolaptr (commented):
Great job!

app/src/main.c
DNedic: That's a good point, this is a problem because the boot count is used in the log filename and in the case of this failing, log file 0 would be overwritten.

DNedic: We could use time and date once we have GNSS or ground control two-way communication.

DNedic: I have reworked this to log to a temporary file when we don't have boot count.

lkijanovic (approved):

app/src/main.c
```diff
@@ -63,7 +63,7 @@ struct fs_mount_t main_fs_mount = {
     .mnt_point = "/SD:",
 };
 
-uint32_t boot_count = 0;
+int32_t boot_count = -1;
```
mileticveljko, line 66: Why didn't you change type to `int32_t` in `logger.c` as well?

DNedic: Ouch, I totally missed that!

DNedic:
Thanks for reviewing this @nikolaptr @lkijanovic @mileticveljko !

nikolaptr (approved):

PR #57, opened 2025-02-15
feat(icm42688p): Added support for ICM42688P trigger

nikolaptr:
Added support for ICM42688P data acquisition via trigger.

### DIFF
--- a/app/src/main.c
+++ b/app/src/main.c
@@ -148,6 +148,13 @@ static int process_imu(const struct device *dev) {
     return 0;
 }
 
+#ifdef CONFIG_ICM42688_TRIGGER
+static void handle_icm42688p_drdy(const struct device *dev,
+                                  const struct sensor_trigger *trig) {
+    process_imu(dev);
+}
+#endif
+
 static int process_baro(const struct device *dev) {
     int ret = sensor_sample_fetch(dev);
     if (ret < 0) {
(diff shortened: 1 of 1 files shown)


app/src/main.c
```diff
@@ -306,7 +329,8 @@ int main(void) {
             return 0;
         }
 
-        process_imu(main_imu);
+        if (!main_imu_trigger)
```
DNedic, line 332: If I am reading into this correctly, you can omit this variable entirely and use `#ifndef CONFIG_ICM42688_TRIGGER`, unless the intention is for polling to be used when configuring the trigger fails in runtime.

app/src/main.c
```diff
@@ -210,12 +217,28 @@ int main(void) {
         return 0;
     }
 
+    uint8_t main_imu_trigger = 0;
```
DNedic, line 220: Can we use a clearer name and a boolean here? Something like `bool main_imu_using_triggers/interrupts` or similar.

app/src/main.c
nikolaptr: That way `#ifndef` would need to take into consideration each supported sensor, which is not ideal (`CONFIG_ICM42688P_TRIGGER`, `CONFIG_MPU6050_TRIGGER`, etc.).

Also, this would also leave room for mistake on the user side. For example, if MPU6050 polling is desired, enabling trigger for any other sensor using `menuconfig` would prevent MPU6050 polling.

DNedic: I see, best keep it this way then, just consider the other comment please.

lkijanovic (approved):

app/src/main.c
```diff
@@ -210,12 +217,28 @@ int main(void) {
...
     const struct device *const main_imu =
         DEVICE_DT_GET(DT_NODELABEL(imu_icm42688p));
+
+#ifdef CONFIG_ICM42688_TRIGGER
+    struct sensor_trigger icm42688p_trigger = {.type = SENSOR_TRIG_DATA_READY,
+                                               .chan = SENSOR_CHAN_ALL};
+
+    ret =
+        sensor_trigger_set(main_imu, &icm42688p_trigger, handle_icm42688p_drdy);
+    if (ret < 0) {
```
DNedic, line 235: Sorry for only seeing this now, but if setting the trigger fails, we could simply not set `main_imu_using_trigger` and try using polling instead instead of terminating.

DNedic (changes requested):

app/src/main.c
nikolaptr: This is an interesting point, but i think we would need to first discuss and implement fallback procedures. Things such as retries, if polling mode is desired substitute, etc...

Specifically, in this case, if `CONFIG_ICM42688_TRIGGER` is defined, sensor will be configured to generate trigger and likewise MCU will be prepared to process this trigger (via `GLOBAL_THREAD` or `OWN_THREAD`). Then, `sensor_trigger_set` will configure the handler and read sensor `INT_STATUS` register, in case status register read fails error will be returned (if the trigger config is valid), but handler will remain configured! Meaning that the `GLOBAL_THREAD` (or `OWN_THREAD`) will continue to call handler.

If we fall back to polling this would result in duplication of data (in best case scenario). So to ensure proper device operation it would be best to reinitialize the device in polling mode, and i can't see a clean way to do this.

nikolaptr: I would suggest to decide on fallback procedure and revisit this after

DNedic: Good points, I agree. Can you create an issue to further discuss fallback and retry mechanisms?

DNedic (approved):

PR #62, opened 2025-02-22
Add ESC support and demo

filipembedded:
This library is written for electronic speed controllers (ESCs) used to drive brushless motors. It is built on top of Zephyr's PWM API and supports popular analog ESC protocols:

- PWM
- Oneshot125
- Oneshot42
- Multishot

The library has been tested on real hardware using T-Motor brushless motors and a BLHeliS electronic speed controller. During my first test with this setup, I noticed a dead zone in certain thrust percentages. What do I mean by that? The motors start spinning only when the thrust reaches 30%. This behavior occurs across all the supported ESC protocols.

I followed the standard pulse durations for these protocols:

- 1000–2000 µs for PWM
- 125–250 µs for Oneshot125
- 42–84 µs for Oneshot42
- 5–25 µs for Multishot

However, the motors do not start spinning when the minimum thrust is applied. I resolved this by applying offsets to the minimum pulse duration values for all analog ESC protocols. While this serves as a quick fix, this unusual behavior should be investigated further.

The usage of this library can be found in demos/esc.


libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+#include <zephyr/drivers/pwm.h>
+#include <zephyr/kernel.h>
+
+/*********************************** Defines **********************************/
+#define MIN_THROTTLE_OFFSET_ON 1
+
+/* NOTE: Manually applied offset bellow is the placeholder for future
+implementation of ESC parameters being stored in NVS (Non Volatile Storage).
+Minimum throttle values can vary from vendor to vendor. In future, it will be
+user-configurable according to used ESCss */
```
nikolaptr, line 38: Having this placeholder is a bit confusing, if configuration of parameters (`DC_MIN_PWM_US` etc.) is needed now it can be done using KConfig, not by changing defines in .h file.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+#define ARM_DC_PWM_US 1000
+#define ARM_DC_ONESHOT_125_US 125
+#define ARM_DC_ONESHOT_42_US 42
+#define ARM_DC_MULTISHOT_US 5
+#define ARM_DURATION 3000 // NOTE: Arm duration is around 2000-5000 ms
+
+/*********************************** Macros ***********************************/
+
+/*********************************** Typedefs *********************************/
+typedef int status_t;
```
nikolaptr, line 87: This can be replaced by enum (i.e. `ESC_Error_Type`) which can further specify return of the functions (i.e. `ESC_SUCCESS`, `ESC_INVALID_PARAM`, `ESC_PWM_ERROR`, etc.)

You can refer to ulog library

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+
+/********************************** Functions *********************************/
+/**
+ * @brief ESC Initialization function
+ * @param[in] pwm_dev     Pointer to the device struct - pwm device
+ * @param[in] pwm_channel PWM channel which will be used by ESC device
+ * @param[in] protocol    ESC Protocol
+ * @param[out] esc_out    Pointer to the ESC out struct, it needs to be passed
+ *                        to the ESC_SetSpeed function
+ * @retval ::status_t
```
nikolaptr, line 136: Return values should be better documented, having only `status_t` as a return doesn't give a lot of information

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+            return -1;
+        }
+
+        k_msleep(ARM_DURATION);
+
+        break;
+
+    default:
+        return -1;
+    }
```
nikolaptr, line 213: As i can see, only difference between these cases is the duty cycle for arm

Switch can be used to set value of some variable (i.e. `arm_duty_cycle`), then `pwm_set`, return check and `k_msleep` can be moved out of the switch as it is common for each protocol. This would make code a lot more readable.

nikolaptr (commented):
Good job! It just needs some refinement

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+/*********************************** Macros ***********************************/
+
+/*********************************** Typedefs *********************************/
+typedef int status_t;
+typedef uint8_t esc_number_t;
+typedef uint32_t esc_period_t;
+typedef uint32_t esc_dc_min_t, esc_dc_max_t;
+
+/**
+ * @brief ESC Protocol type
```
DNedic, line 93: Nitpick: These comments seem pretty useless considering they just repeat what is already spelled out in the type name.

libs/motor_control/esc/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+set(SOURCES 
+    esc.c    
+)
+
+# Create the library
+add_library(${PROJECT_NAME} STATIC ${SOURCES})
+
+target_include_directories(${PROJECT_NAME}
+SYSTEM PUBLIC
```
DNedic, line 30: Why are we including the `esc` header as a system header? This would suppress any warnings or errors originating from a header we control, and is not a system header fundamentally.

libs/motor_control/esc/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${ZEPHYR_BASE}/include
```
DNedic, line 32: Never include another targets folders directly, you should link to a target interface which includes them, and we are doing this infact 3 lines down.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * @brief ESC Initialization function
+ * @param[in] pwm_dev     Pointer to the device struct - pwm device
+ * @param[in] pwm_channel PWM channel which will be used by ESC device
+ * @param[in] protocol    ESC Protocol
+ * @param[out] esc_out    Pointer to the ESC out struct, it needs to be passed
+ *                        to the ESC_SetSpeed function
+ * @retval ::status_t
+ */
+status_t ESC_Init(const struct device *pwm_dev, uint32_t pwm_channel,
```
DNedic, line 138: Nitpick: Perhaps it would be more inuitive to take an inst pointer as a first argument like in all other functions?

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+
+    return 0;
+}
+
+status_t ESC_SetSpeed(esc_t *esc, uint8_t speed) {
+    int ret;
+    uint32_t dc_speed;
+
+    if (esc->flag != ESC_INITIALIZED) {
+        printk("[ESC]: Error: Uninitialized ESC cannot be started!");
```
DNedic, line 95: As @nikolaptr has already mentioned, we do not need these, but in cases where you do want to print errors in a library, please use Zephyrs logging subsystem.

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        k_msleep(ARM_DURATION);
+
+        break;
+
+    case ESC_MULTISHOT:
+        ret = pwm_set(esc->instance.pwm_dev, esc->instance.pwm_channel,
+                      PWM_USEC(esc->period), PWM_USEC(ARM_DC_MULTISHOT_US), 0);
+        if (ret) {
```
DNedic, line 200: These `if (ret)` segments can be extracted out of the switch statement since all cases call the same function.

DNedic (changes requested):
Thank you for working on this! I've left a lot of comments, please take your time to address them.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+#define DC_MIN_ONESHOT_125_US 162.5
+#define DC_MAX_ONESHOT_125_US 250
+
+#define DC_MIN_ONESHOT_42_US 54.6
+#define DC_MAX_ONESHOT_42_US 84
+
+#define DC_MIN_MULTISHOT_US 11
+#define DC_MAX_MULTISHOT_US 25
+
+#else
```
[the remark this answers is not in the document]
lkijanovic: And if they don't they shouldn't be in the publicly exposed header file.

demos/esc/Kconfig
```diff
@@ -0,0 +1,10 @@
+mainmenu "Eurus Flight Controller"
+
+config APP_DATA_LOGGING_SYNC_INTERVAL
+	int "Data logging Sync interval [ms]"
+	default 1000
+	help
+		This configures the interval at which the logging system flushes everything to disk
+		and hence determines the maximum duration of data lost on power cuts.
+
+source "Kconfig.zephyr"
```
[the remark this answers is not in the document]
filipembedded: Good point, thanks!

demos/esc/src/main.c
```diff
@@ -0,0 +1,128 @@
...
+    /* ESC Initialization */
+    status_t status;
+    esc_t esc1;
+
+    const struct device *pwm_dev = DEVICE_DT_GET(DT_NODELABEL(pwm1));
+
+    if (!device_is_ready(pwm_dev))
+        return;
+
+    status = ESC_Init(pwm_dev, 1, ESC_PWM, &esc1);
```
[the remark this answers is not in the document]
filipembedded: I agree with this. But at the moment of writing this library I didn't think about this, functionality was my man goal. I will change this, thanks!

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+
+    default:
+        return -1;
+    }
+
+    return 0;
+}
+
+/************************** Static Functions **********************************/
+static int prvESC_ChannelCheck(const struct device *pwm_dev,
```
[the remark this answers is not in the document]
filipembedded: Okay, I will open sub-issue for this check.

libs/motor_control/esc/esc.h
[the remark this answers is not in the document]
filipembedded: The naming of this defines confused me at the moment of writing this. This defines does not need to be seen by an upper layer definitely, they should be fixed, as they are by the universal documentation related to the ESC protocols. But, the applied offset, that I hardcoded, should be in future solved by pulling the minimal pulse values (DC_MIN_) from some Non-volatile memory storage, values that had been configured by the user previously (these values can differ from ESC to ESC, and are directly tied to the ESC vendor and ESC firmware) 
For now, I can replace  MIN_THROTTLE_OFFSET_ON define with some KConfig option that can be set by an user, what is your opinion about this?

libs/motor_control/esc/esc.h
filipembedded: I understand!

libs/motor_control/esc/esc.c
filipembedded: You are right! I will update this

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file   esc.c
+ * @brief  Zephyr Electronic Speed Controller ESC driver for BLDC motors
+ * @author Eurus-project
+ */
+
+/******************************** Includes ************************************/
```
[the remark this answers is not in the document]
filipembedded: No problem, I will change this!

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,222 @@
...
+        return -1;
+    }
+
+    return 0;
+}
+
+/************************** Static Functions **********************************/
+static int prvESC_ChannelCheck(const struct device *pwm_dev,
+                               uint32_t pwm_channel) {
+    // TODO: Implement this in future!
```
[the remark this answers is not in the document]
filipembedded: Understandable, I will follow this approach in the future. Thank you!

libs/motor_control/esc/esc.h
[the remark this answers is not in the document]
filipembedded: This was just a placeholder from the past. I didn't know if we would use enums as a return types or some standard return codes. @nikolaptr pointed this, I will change it. Thank you!

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+                  esc_protocol_t protocol, esc_t *esc_out);
+
+/**
+ * @brief ESC Set Speed by percentage from 0 to 100
+ * @param[in] esc   Pointer to the preinitialized ESC out struct from ESC_Init
+ * @param[in] speed ESC speed, in perentage
+ *
+ * @retval ::status_t
+ */
+status_t ESC_SetSpeed(esc_t *esc, uint8_t speed);
```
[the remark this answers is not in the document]
filipembedded: I don't think we should worry about this at the moment. As my understanding of multicopters goes, only motors with one spinning direction has been used. There is no need to change direction of motor spinning. You made a very good point, and I will definitely consider adding this functionality in future versions of ESC library.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,173 @@
...
+#ifndef ESC_H
+#define ESC_H
+
+/********************************** Includes **********************************/
+#include <stdio.h>
+#include <zephyr/drivers/pwm.h>
+#include <zephyr/kernel.h>
+
+/*********************************** Defines **********************************/
+#define MIN_THROTTLE_OFFSET_ON 1
```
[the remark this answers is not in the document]
filipembedded: Good point! I will add the KConfig option for this, as @nikolaptr proposed

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,172 @@
...
+
+#define ESC_APPLY_MIN_OFFSET(OFFSET, DC_MIN, DC_MAX)                           \
+    ((DC_MIN) + (((DC_MAX) - (DC_MIN)) * ((OFFSET) / 100.0f)))
+#define ESC_APPLY_MAX_OFFSET(OFFSET, DC_MIN, DC_MAX)                           \
+    ((DC_MAX) - (((DC_MAX) - (DC_MIN)) * ((OFFSET) / 100.0f)))
+
+// Defined in Kconfig
+#if CONFIG_ESC_MANUAL_MIN_THRUST_OFFSET
+
+#define ESC_DC_MIN_PWM_US                                                      \
```
DNedic, line 33: Defines like these offer no type-safety, use functions instead.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,172 @@
...
+
+typedef struct {
+    uint8_t instance_count;
+    const struct device *pwm_dev;
+    uint32_t pwm_channel;
+    uint32_t period;
+    float dc_min;
+    float dc_max;
+    ESC_Protocol_Type protocol;
+    bool flag;
```
DNedic, line 112: Flag is a very unclear variable name, rename this to indicate what it actually represents.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,172 @@
...
+
+typedef enum {
+    ESC_PWM = 0,
+    ESC_ONESHOT_125,
+    ESC_ONESHOT_42,
+    ESC_MULTISHOT
+} ESC_Protocol_Type;
+
+typedef struct {
+    uint8_t instance_count;
```
DNedic, line 105: This is unnecessary IMO, since we're not using this anywhere, nor would I know why we would use it.

demos/esc/prj.conf
```diff
@@ -0,0 +1,39 @@
...
+
+CONFIG_MAIN_STACK_SIZE=4096
+
+CONFIG_LOG=y
+CONFIG_DEBUG_COREDUMP=y
+CONFIG_DEBUG_COREDUMP_BACKEND_LOGGING=y
+
+CONFIG_CBPRINTF_FP_SUPPORT=y
+
+CONFIG_INPUT=y
```
DNedic, line 38: Some of these Kconfig settings are not needed for the demo, for instance this one.

DNedic (changes requested):
Nice work @filipembedded ! This is a big improvement over the last review round. Please address things not already addressed from it and the comments I have left now and this should be ready to merge.

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,164 @@
...
+    uint32_t dc;
+
+    if (!esc->flag) {
+        return ESC_NOT_INITIALIZED;
+    }
+
+    if (speed > 1.0f || speed < 0.0f)
+        return ESC_INVALID_SPEED_SET;
+
+    dc = (uint32_t)(ESC_US_TO_NS(((esc->dc_max - esc->dc_min) * speed) +
```
DNedic, line 86: Again, this is not duty cycle, but pulse duration, also the defines are not really needed and it would be much better to just create an intermediate variable like `pulse_duration_us` and then convert it accordingly later, either in the function call body or with another intermediate variable like `pulse_duration_ns`.

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,172 @@
...
+    const struct device *pwm_dev;
+    uint32_t pwm_channel;
+    uint32_t period;
+    float dc_min;
```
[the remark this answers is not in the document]
filipembedded: This is a good suggestion, thanks!

libs/motor_control/esc/esc.h
```diff
@@ -0,0 +1,172 @@
...
+#include <zephyr/drivers/pwm.h>
+#include <zephyr/kernel.h>
+
+#define ESC_APPLY_MIN_OFFSET(OFFSET, DC_MIN, DC_MAX)                           \
+    ((DC_MIN) + (((DC_MAX) - (DC_MIN)) * ((OFFSET) / 100.0f)))
+#define ESC_APPLY_MAX_OFFSET(OFFSET, DC_MIN, DC_MAX)                           \
+    ((DC_MAX) - (((DC_MAX) - (DC_MIN)) * ((OFFSET) / 100.0f)))
+
+// Defined in Kconfig
+#if CONFIG_ESC_MANUAL_MIN_THRUST_OFFSET
```
DNedic, line 31: This is defined in the demo Kconfig but you're checking for it in the library, this won't exist when the library is used outside of the demo and also it is very unintuitive to depend on app Kconfig for a library, if configuring a library create a Kconfig for that library.

demos/esc/Kconfig
```diff
@@ -0,0 +1,37 @@
+mainmenu "Eurus Flight Controller - ESC demo"
+
+config ESC_MANUAL_MIN_THRUST_OFFSET
```
DNedic, line 3: App Kconfig variables should be prefixed with APP.

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,227 @@
...
+        ESC_Stop(esc);
+        return ESC_INVALID_ARMING;
+    }
+
+    k_msleep(ESC_ARM_DURATION);
+
+    return ESC_SUCCESS;
+}
+
+static float ESC_ApplyMinOffset(float min_pulse_val_us, float max_pulse_val_us,
```
DNedic, line 217: Nitpick: private functions do not need to be namespaced.

demos/esc/Kconfig
```diff
@@ -0,0 +1,25 @@
...
+
+config ESC_ONESHOT_42
+    bool "ESC Oneshot42 protocol"
+
+config ESC_MULTISHOT
+    bool "ESC Multishot protocol"
+
+endchoice
+
+source "../efc/libs/motor_control/esc/Kconfig"
```
DNedic, line 25: If the intent is for us to use this method of sourcing Kconfig files for the time being, this should also be present in the main firmware.

Long-term I guess we want to use a more reliable method (perhaps Zephyr modules), since it is easy to forget to do this and we shouldn't need to both add a library in CMake and source Kconfig as separate steps.

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,227 @@
...
+    case ESC_PWM:
+        esc_out->period_us = ESC_T_PWM_US;
+#ifdef CONFIG_ESC_MANUAL_MIN_PULSE_OFFSET
+        esc_out->min_pulse_duration_us =
+            ESC_ApplyMinOffset(ESC_MIN_PULSE_PWM_US, ESC_MAX_PULSE_PWM_US,
+                               CONFIG_ESC_MIN_PULSE_OFFSET_PERCENT);
+#else
+        esc_out->min_pulse_duration_us = ESC_MIN_PULSE_PWM_US;
+#endif
+#ifdef CONFIG_ESC_MANUAL_MAX_PULSE_OFFSET
```
DNedic, line 48: Can we find a more programatic way to do min/max pulse duration setting? A switch full of #ifdefs like this seems difficult to maintain.

libs/motor_control/esc/esc.c
```diff
@@ -0,0 +1,227 @@
...
+        return ESC_NOT_INITIALIZED;
+    }
+
+    ESC_Stop(esc);
+    esc->initialized = false;
+
+    return ESC_SUCCESS;
+}
+
+ESC_Error_Type ESC_Arm(ESC_Inst_Type *esc) {
```
DNedic, line 176: Arm is a weird way to name this since it actually starts outputting a PWM signal, `Start` would make more sense IMO.

libs/motor_control/esc/Kconfig
```diff
@@ -0,0 +1,31 @@
...
+
+config ESC_MAX_PULSE_OFFSET_PERCENT
+    int "Duty cycle offset in percentage"
+    default 0
+    depends on ESC_MANUAL_MAX_PULSE_OFFSET
+
+
+endif # ESC
+
+source "Kconfig.zephyr"
```
DNedic, line 31: Are we sure we need this here?

libs/motor_control/esc/esc.c
filipembedded: This is the Arming procedure, I don't see the issue here. I think that naming it ESC_Start would definitely confuse some users of this library. "Start" of ESCs is never used in UAV world as a term.

libs/motor_control/esc/esc.c
filipembedded: But do we need any prefix, something like "prv", in example?

libs/motor_control/esc/esc.c
filipembedded: I agree, it is a little bit messy. But I didn't have any cleaner idea at the time.

libs/motor_control/esc/Kconfig
filipembedded: Do you mean **source "Kconfig.zephyr"**? I am not sure about this, I need to check it.

libs/motor_control/esc/esc.c
DNedic: No need for that either IMO, it should be clear that something is private from the fact that it isn't namespaced.

libs/motor_control/esc/Kconfig
filipembedded: @DNedic I have checked this. **source "Kconfig.zephyr"** is absolutely needed here, when I omit it, a lot of errors pop up - errors in arch_v2.cmake, zephyr_default.cmake, ZephyrConfig.cmake, ...

demos/esc/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Features & Demo configuration
+
+The library supports multiple ESC protocols, including:
+- Classic PWM
+- Oneshot 125
+- Oneshot 42
+- Multishot
+
+Default ESC protocol in this sample is PWM, but it can be reconfigured 
+using menuconfig. To start menuconfig, the demo needs to be built with:
```
DNedic, line 16: In markdown, there is no need to wrap lines around a certain character count, the viewer will do that. Generally it is best to start new lines for new sentences and double newlines for paragraphs.

demos/esc/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Oneshot 125
+- Oneshot 42
+- Multishot
+
+Default ESC protocol in this sample is PWM, but it can be reconfigured 
+using menuconfig. To start menuconfig, the demo needs to be built with: 
+
+`west build -b <board_name> <path_to_demo>`
+
+then, to start menuconfig enter:
```
DNedic, line 20: Order of operations should be reversed, menuconfig should be run first if the user wants to configure something.

demos/esc/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+`west build -t menuconfig`
+
+![Eurus Flight Controller - ESC Demo menuconfig](.docs/menuconfig1.png)
+
+Next, by choosing ESC Protocol Selection, submenu will open where ESC PWM
```
DNedic, line 26: We don't need to teach people how to use menuconfig, this entire section can be removed. It's enough to list available protocols and variables and how to enter menuconfig.

demos/esc/README.md
```diff
@@ -0,0 +1,85 @@
...
+![ESC Library settings](.docs/menuconfig4.png)
+
+All analog ESC protocols have a standard minimal and maximal pulse durations:
+- PWM        -> 1000-2000 us
+- Oneshot125 -> 125-250 us
+- Oneshot42  -> 42-84 us
+- Multishot  -> 5-25 us
+
+But in reality, this values can vary from vendor to vendor. In this demo, ESC 
+that has been used is BLHeliS ESC, and it has a dead zone of 30 % starting from
```
DNedic, line 46: Nitpick: A link to the manufacturers page of the mentioned ESC would be nice.

demos/esc/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+![Enable manual minimum thrust offset for ESC](.docs/menuconfig5.png)
+
+**Enable manual maximum thrust offset for ESC** can be used to set offset of maximum pulse duration, but in this demo it wasn't necessary.
+
+## Demo
+
+### Requirements
+
+#### Hardware
```
DNedic, line 59: Nitpick: There is no need to have this many levels of headers, the same effect can be achieved with just one or two.

demos/esc/README.md
filipembedded: That seemed logical to me too, but I have noticed that if there is no build directory (for example user is configuring this project for the first time) and I run command `west build -t menuconfig` it throws me a warning: `WARNING: This looks like a fresh build and BOARD is unknown; so it probably won't work. To fix, use --board=<your-board>.`, and an error: `ninja: error: unknown target 'menuconfig'` And to solve it, I tried to run the command `west build -b <board_name> -t menuconfig` , `west build -t menuconfig -b <board_name>` and `west build -t menuconfig --board=<board_name>` and nothing worked. Only way that worked is the one I wrote in Readme. Any suggestions here?

libs/motor_control/esc/esc.h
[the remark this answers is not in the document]
DNedic: Please move these into the source file for now, they shouldn't be publicly exposed. We can explore options for customization in a follow-up PR since this one was in review for a really long time and its about time to merge it.

DNedic (approved):
LGTM, huge improvement over what we started with!

mileticveljko (approved):
Great job!

PR #63, opened 2025-02-24
feat: Add shell support and enable fs commands

### LINKED ISSUE #61: Add a convenient way to pull logs
Right now, the only way to obtain logs is to pull the SD card out of the autopilot, hook it up to the host PC with an SDC reader and use `littlefs-fuse` to read its contents. This is not only a chore to do, but also does not leave any room for changing the filesystem type or moving to a different storage medium.

By implementing an easy way for the host to obtain the logs from efc itself, we would save ourselves a lot of time and create a future-proof mechanism.

DNedic:
This enables shell support on the USB CDC ACM serial connection and turns on the filesystem shell commands in order to be able to obtain the logs without pulling the SD card out. Down the line other shell commands will be added as well, for instance for ESC calibration or parameter changes.

As a notable downside of the logging backend being used for the shell, logging messages will not be printed before the shell thread is running, which might obscure early bootup issues, but these can be handled on a case-by-case basis IMO.

Closes #61


lkijanovic (approved):
Not much to review, pretty straightforward, good job.
Also, two minor questions:
- There aren't any conflicts between console and shell as they are using same UART?
- Is there any additional error-checking mechanism when e.g. reading logs (other than UART parity bit)? I think transfers of large log files could be quite error-prone.

DNedic:
> Not much to review, pretty straightforward, good job. Also, two minor questions:
> 
> * There aren't any conflicts between console and shell as they are using same UART?

There are no conflicts, the two subsystems work together, the shell prompt is re-printed after log messages and also any continous shell output is guaranteed not to be split by logging messages.

> * Is there any additional error-checking mechanism when e.g. reading logs (other than UART parity bit)? I think transfers of large log files could be quite error-prone.

At the moment, no, however for the time being we are using this interface over USB which has parity checks embedded in every packet. In the future we could drop `fs` commands and use our own shell commands which implement a CRC for files.

nikolaptr (approved):
LGTM, great job!

DNedic:
Thanks for reviews @lkijanovic @nikolaptr !

PR #65, opened 2025-03-07
docs: Add contributing note regarding unit variables

DNedic:
This makes sure that there is no confusion when doing math with physical unit based variables.


filipembedded (approved):
Good suggestion! I think we should stick to that approach in the future.

PR #66, opened 2025-04-26
esc: Change outdated error status handling

filipembedded:
Previous error handling used a raw `if (status < 0)` check, which is
no longer appropriate after migrating to standardized error codes.
Remove the outdated check to align with the updated error handling
approach.


DNedic (approved):
LGTM

DNedic:
Please note that this shouldn't be labeled `feat`, since its a fix and not a feature and also not under `esc` since its a part of the demos, not the esc library.

PR #67, opened 2025-05-19
fix: Consolidate CMake versions across efc

lkijanovic:
CMake >= 4.0 fails to build as it now requires cmake_minimum_required() version >= 3.5.0.
Zephyr and MAVLink satisfy this, and this is the only occurrence in our repository.

Even without new limitations it would be nice to keep consistent with stuff like this across non-external files.


DNedic (approved):
LGTM

filipembedded (approved):

PR #68, opened 2025-07-19
boards(eurus_nexus_v1_1): Add initial support

filipembedded:
This adds initial support for a custom board - eurus_nexus_v1_1.


boards/eurus_project/eurus_nexus_v1_1/doc/eurus_nexus_v1_1_schematic.pdf
DNedic, line 1: Could you check the terms of conditions of EasyEDA for inclusion here?

app/boards/eurus_nexus_v1_1.overlay
DNedic, line 1: We don't *really* need an overlay, since this is a single-purpose board, and has most things fixed for this application specifically.

boards/eurus_project/eurus_nexus_v1_1/eurus_nexus_v1_1_defconfig
DNedic, line 1: Do we have everything we need here? `prj.conf` we are using by default could mask missing things here.

DNedic (approved):
LGTM overall, very nice work!

boards/eurus_project/eurus_nexus_v1_1/eurus_nexus_v1_1_defconfig
filipembedded: That makes sense! I will check it out

app/boards/eurus_nexus_v1_1.overlay
filipembedded: I agree. But at this moment, a couple of nodes defined in devicetree are not being used in our main application, so I used the overlay to disable unused nodes.

boards/eurus_project/eurus_nexus_v1_1/eurus_nexus_v1_1_defconfig
```diff
@@ -0,0 +1,61 @@
...
+# 
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+CONFIG_GPIO=y
+CONFIG_SERIAL=y
+CONFIG_CONSOLE=y
+CONFIG_USB_DEVICE_STACK=y
+CONFIG_I2C=y
```
DNedic, line 22: Everything under here can be removed as it is not board-specific.

PR #71, opened 2025-10-06
demos(nvs-nor-flash): Add demo for Zephyr Non-Volatile Storage (NVS) on external flash

filipembedded:
This PR adds adapted version of the [Zephyr NVS demo](https://docs.zephyrproject.org/latest/samples/subsys/nvs/README.html) tailored to the needs of **efc** project.

### DIFF
--- a/demos/nvs-spi-flash/Kconfig
+++ b/demos/nvs-spi-flash/Kconfig
@@ -1,0 +1,3 @@
+# Flash non-volatile storage demo configuration
+
+source "Kconfig.zephyr"
--- a/demos/nvs-spi-flash/prj.conf
+++ b/demos/nvs-spi-flash/prj.conf
@@ -1,0 +1,32 @@
+# 
+# This file is part of the efc project <https://github.com/eurus-project/efc/>.
+# Copyright (c) 2024 The efc developers.
+# 
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, version 3.
+# 
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# General Public License for more details.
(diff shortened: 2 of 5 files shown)


demos/nvs-spi-flash/Kconfig
```diff
@@ -0,0 +1,27 @@
+# Flash non-volatile storage demo configuration
+
+mainmenu "NVS Sample Configuration"
+
+config NVS_SAMPLE_MAX_REBOOT
```
DNedic, line 5: Let's not use `NVS_` for app-level configs, as they can be confused with actual NVS subsystem configuration values. Either `DEMO_` or `APP_` would suffice.

demos/nvs-spi-flash/prj.conf
```diff
@@ -0,0 +1,34 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+CONFIG_FLASH=y
+CONFIG_NVS=y
+CONFIG_LOG=y
+CONFIG_LOG_MODE_IMMEDIATE=y
+CONFIG_NVS_LOG_LEVEL_DBG=y
+CONFIG_REBOOT=y
+#CONFIG_MPU_ALLOW_FLASH_WRITE=y #If NVS is used with INTERNAL flash, enable this
```
DNedic, line 24: Let's not leave commented-in code.

demos/nvs-spi-flash/src/main.c
```diff
@@ -0,0 +1,268 @@
+/*
+ * NOTE: This demo is the official Zephyr demo adapted to suit the needs of efc
```
DNedic, line 2: We're not allowed to do this. `efc` is licensed under GPLv3, and the Zephyr sample is under Apache 2.0. You're not allowed to apply a more restrictive license to a piece of code AFAIK.

DNedic: As an alternative, why not suggest building the official sample with our board?

If that is not an option, we have to implement the sample from scratch, without looking at the Zephyr sample.

PR #81, opened 2026-01-03
Add motor mixing library and demo

filipembedded:
This PR adds motor mixing library for quadcopters and demo that showcases the usage of the library.

### DIFF
--- a/libs/motor_control/esc/Kconfig
+++ b/libs/motor_control/esc/Kconfig
@@ -27,6 +27,3 @@ config ESC_MAX_PULSE_OFFSET_PERCENT
 
 
 endif # ESC
-
-
-source "Kconfig.zephyr"
--- a/libs/motor_control/mixer/mixer.c
+++ b/libs/motor_control/mixer/mixer.c
@@ -42,18 +42,19 @@ MIXER_Error_Type MIXER_AddMotorInstance(MIXER_Inst_Type *mixer,
         return MIXER_ESC_ERROR;
     }
 
-    mixer->motor_arr[mixer->motor_instances] = *esc;
-    mixer->motor_instances++;
-
-    if (mixer->motor_instances > mixer->max_motor_num) {
+    if (mixer->motor_instances == mixer->max_motor_num) {
         return MIXER_INIT_ERROR;
     }
+
+    mixer->motor_arr[mixer->motor_instances] = *esc;
+    mixer->motor_instances++;
 
     return MIXER_SUCCESS;
 }
 
 MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
                             MIXER_UAV_Cfg_Type uav_cfg) {
+    MIXER_Error_Type ret = MIXER_SUCCESS;
 
     if (mixer == NULL) {
         return MIXER_INIT_ERROR;
@@ -67,8 +68,6 @@ MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
     mixer->uav_config = uav_cfg;
     mixer->motor_instances = 0;
 
-    MIXER_Error_Type ret = MIXER_SUCCESS;
-
     switch (uav_cfg) {
     case MIXER_UAV_CFG_QUADROTOR_X:
     case MIXER_UAV_CFG_QUADROTOR_CROSS:
@@ -80,11 +79,15 @@ MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
         break;
     default:
         ret = MIXER_INVALID_CFG;
-        mixer->initialized = false;
         break;
     }
 
-    mixer->initialized = true;
+    if (ret == MIXER_SUCCESS) {
+        mixer->initialized = true;
+    } else {
+        mixer->initialized = false;
+    }
+
     return ret;
 }
 
--- a/libs/motor_control/mixer/mixer.c
+++ b/libs/motor_control/mixer/mixer.c
@@ -120,26 +120,25 @@ static void normalize_stick_input(const MIXER_Raw_Input_Type *mixer_raw,
 #endif
 
 #ifdef CONFIG_MIXER_MAXIMAL_RECEIVER_DATA_OFFSET
-    max_receiver_val = CONFIG_MIXER_MAXIMAL_RECEIVER_DATA_VALUE
+    max_receiver_val = CONFIG_MIXER_MAXIMAL_RECEIVER_DATA_VALUE;
 #else
     max_receiver_val = MIXER_RECEIVER_DEFAULT_MAX_VALUE;
 #endif
 
 #ifdef CONFIG_MIXER_NEUTRAL_RECEIVER_DATA_OFFSET
-        neutral_receiver_val = CONFIG_MIXER_NEUTRAL_RECEIVER_DATA_VALUE
+    neutral_receiver_val = CONFIG_MIXER_NEUTRAL_RECEIVER_DATA_VALUE;
 #else
     neutral_receiver_val = MIXER_RECEIVER_DEFAULT_NEUTRAL_VALUE;
 #endif
 
-        /* Map raw roll values:
-            [min_receiver_val, neutral_receiver_val] -> [-1.0, 0.0] and
-            [neutral_receiver_val, max_receiver_val] -> [0.0, 1.0]
-        */
-        if (mixer_raw->roll < neutral_receiver_val) {
+    /* Map raw roll values:
+        [min_receiver_val, neutral_receiver_val] -> [-1.0, 0.0] and
+        [neutral_receiver_val, max_receiver_val] -> [0.0, 1.0]
+    */
+    if (mixer_raw->roll < neutral_receiver_val) {
         mixer_mapped->roll = (float)(mixer_raw->roll - neutral_receiver_val) /
                              (float)(neutral_receiver_val - min_receiver_val);
-    }
-    else {
+    } else {
         mixer_mapped->roll = (float)(mixer_raw->roll - neutral_receiver_val) /
                              (float)(max_receiver_val - neutral_receiver_val);
     }
--- a/libs/motor_control/mixer/mixer.c
+++ b/libs/motor_control/mixer/mixer.c
@@ -17,16 +17,16 @@
 
 #include "mixer.h"
 #include <stdio.h>
+#include <string.h>
 #include <zephyr/kernel.h>
 
-#define MIXER_RECEIVER_DEFAULT_MIN_VALUE 0.0f
-#define MIXER_RECEIVER_DEFAULT_MAX_VALUE 2047.0f
-#define MIXER_RECEIVER_DEFAULT_NEUTRAL_VALUE 1024.0f
-
-static void normalize_stick_input(const MIXER_Raw_Input_Type *mixer_raw,
-                                  MIXER_Mapped_Input_Type *mixer_mapped);
-
-static float limit_min_thrust_values(float mixer_calcualated_val, float thrust);
+static void setup_motor_factors(MIXER_Inst_Type *mixer);
+static MIXER_Error_Type validate_mixer_input(const MIXER_Input_Type *input);
+static void calculate_motor_outputs(MIXER_Inst_Type *mixer,
+                                    const MIXER_Input_Type *input);
+static void normalize_motor_outputs(MIXER_Inst_Type *mixer);
+static MIXER_Error_Type apply_motor_outputs(MIXER_Inst_Type *mixer);
+static MIXER_Error_Type validate_mixer_geom_cfg(MIXER_UAV_Cfg_Type cfg);
 
 MIXER_Error_Type MIXER_AddMotor(MIXER_Inst_Type *mixer, ESC_Inst_Type *esc) {
     if (mixer == NULL || esc == NULL) {
@@ -59,13 +59,17 @@ MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
         return MIXER_INIT_ERROR;
     }
 
-    if (uav_cfg < MIXER_UAV_CFG_QUADROTOR_X ||
-        uav_cfg > MIXER_UAV_CFG_HEXAROTOR_CROSS) {
+    if (validate_mixer_geom_cfg(uav_cfg) != MIXER_SUCCESS) {
         return MIXER_INVALID_CFG;
     }
 
-    mixer->uav_config = uav_cfg;
+    memset(mixer->motor_arr, 0, sizeof(mixer->motor_arr));
+    memset(mixer->motor_factors, 0, sizeof(mixer->motor_factors));
+    memset(mixer->motor_outputs, 0, sizeof(mixer->motor_outputs));
     mixer->motor_instances = 0;
+    mixer->max_motor_num = 0;
+    mixer->uav_config = uav_cfg;
+    mixer->initialized = false;
 
     switch (uav_cfg) {
     case MIXER_UAV_CFG_QUADROTOR_X:
@@ -85,6 +89,7 @@ MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
 
     if (ret == MIXER_SUCCESS) {
         mixer->initialized = true;
+        setup_motor_factors(mixer);
     } else {
         mixer->initialized = false;
     }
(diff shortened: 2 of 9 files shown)


libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,240 @@
...
+
+    mixer->motor_arr[mixer->motor_instances] = *esc;
+    mixer->motor_instances++;
+
+    return MIXER_SUCCESS;
+}
+
+MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer,
+                            MIXER_UAV_Cfg_Type uav_cfg) {
+    MIXER_Error_Type ret = MIXER_SUCCESS;
```
DNedic: Never default to success, there are many pitfalls associated, including accidentally returning ret on error, misleading coredumps etc.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,240 @@
...
+                                     mixer_mapped.thrust);
+        m4 = limit_min_thrust_values(mixer_mapped.thrust - mixer_mapped.roll -
+                                         mixer_mapped.pitch - mixer_mapped.yaw,
+                                     mixer_mapped.thrust);
+        break;
+
+    default:
+        break;
+    }
+    esc_status = ESC_SetSpeed(&mixer->motor_arr[0], m1);
```
DNedic: Idea: How about matrix operations from the get-go instead of hardcoding like this?

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,71 @@
...
+    uint8_t max_motor_num;
+} MIXER_Inst_Type;
+
+MIXER_Error_Type MIXER_AddMotorInstance(MIXER_Inst_Type *mixer,
+                                        ESC_Inst_Type *esc);
+
+MIXER_Error_Type MIXER_Init(MIXER_Inst_Type *mixer, MIXER_UAV_Cfg_Type uav_cfg);
+
+MIXER_Error_Type MIXER_Execute(MIXER_Inst_Type *mixer,
+                               MIXER_Raw_Input_Type *mixer_in);
```
DNedic: Why ingest the "raw" data and what format is it in? At least should be documented, ideally should have a less error-prone API like using floats.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,71 @@
...
+
+typedef struct {
+    ESC_Inst_Type motor_arr[MAX_MOTOR_INSTANCES];
+    uint8_t motor_instances;
+    MIXER_UAV_Cfg_Type uav_config;
+    bool initialized;
+    uint8_t max_motor_num;
+} MIXER_Inst_Type;
+
+MIXER_Error_Type MIXER_AddMotorInstance(MIXER_Inst_Type *mixer,
```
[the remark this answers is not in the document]
filipembedded: I agree, that sounds much better.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,240 @@
...
+    if (mixer_mapped->yaw > 1.0f)
+        mixer_mapped->yaw = 1.0f;
+
+    /* Map raw thrust values:
+        [min_receiver_val, max_receiver_val] -> [0.0, 1.0]
+    */
+    mixer_mapped->thrust = (float)(mixer_raw->thrust - min_receiver_val) /
+                           (float)(max_receiver_val - min_receiver_val);
+
+    // Clamp calculated values
```
DNedic: These clampings have the same problem. If one clips, but others don't you get wrong inputs propagated. The input vector should be calculated and normalized.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,240 @@
...
+    ESC_Error_Type esc_status;
+    MIXER_Mapped_Input_Type mixer_mapped;
+
+    /* Remapping the raw values to ESC-compatible ones */
+    normalize_stick_input(mixer_raw, &mixer_mapped);
+
+    // TODO: Add checks here for edge-cases like R=P=0 and Y>T
+    switch (mixer->uav_config) {
+    case MIXER_UAV_CFG_QUADROTOR_X:
+        m1 = limit_min_thrust_values(mixer_mapped.thrust - mixer_mapped.roll +
```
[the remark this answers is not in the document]
filipembedded: I understand the issue here, thanks for pointing this. I think that @nikolaptr mentioned somewhere that this issue could be solved statically, and that it could be simpler solution. 
@nikolaptr Can you point me to some materials or prior implementations related to this issue so I can be sure I approach this correctly?

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,240 @@
...
+
+    return MIXER_SUCCESS;
+}
+
+static void normalize_stick_input(const MIXER_Raw_Input_Type *mixer_raw,
+                                  MIXER_Mapped_Input_Type *mixer_mapped) {
+    int32_t min_receiver_val, max_receiver_val, neutral_receiver_val;
+
+#ifdef CONFIG_MIXER_MINIMAL_RECEIVER_DATA_OFFSET
+    min_receiver_val = CONFIG_MIXER_MINIMAL_RECEIVER_DATA_VALUE;
```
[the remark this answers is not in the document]
filipembedded: That makes sense, thanks a lot!

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,71 @@
...
+
+typedef enum {
+    MIXER_SUCCESS = 0,
+    MIXER_INIT_ERROR,
+    MIXER_ESC_ERROR,
+    MIXER_INVALID_CFG,
+} MIXER_Error_Type;
+
+typedef enum {
+    MIXER_UAV_CFG_QUADROTOR_X = 0,
```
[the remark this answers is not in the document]
filipembedded: Understood! I removed unsupported configurations.

libs/motor_control/mixer/mixer.c
filipembedded: I agree. I implemented matrix operations, please check it out

libs/motor_control/mixer/mixer.c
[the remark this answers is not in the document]
filipembedded: I implemented some solution for saturation based on the way Ardupilot does it. Please check it out @DNedic

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,243 @@
...
+        ESC_Error_Type esc_status =
+            ESC_SetSpeed(&mixer->motor_arr[i], mixer->motor_outputs[i]);
+        if (esc_status < 0) {
+            return MIXER_ESC_ERROR;
+        }
+    }
+    return MIXER_SUCCESS;
+}
+
+static void normalize_motor_outputs(MIXER_Inst_Type *mixer) {
```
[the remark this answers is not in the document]
DNedic: To elaborate further - we want the *euclidean norm*/magnitude of the input vector to be such that we never enter saturation on any one of the motors. You can take a look at edge cases to infer where we would have maximum clipping and then go back from that to the normalization. If we simply go the static route we would find a scaling factor that is always applied, but for dynamic normalization it would be done here on the fly.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,234 @@
...
+}
+
+static void normalize_control(MIXER_Inst_Type *mixer,
+                              MIXER_Input_Type *mixer_in) {
+    switch (mixer->norm) {
+    case MIXER_NORM_STATIC:
+        const float scale = 1.0f / 3.0f;
+        mixer_in->roll *= scale;
+        mixer_in->pitch *= scale;
+        mixer_in->yaw *= scale;
```
lkijanovic: Shouldn't throttle be normalized here as well? It seems that with sufficiently large throttle this will still cause saturation.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,234 @@
...
+        break;
+
+    case MIXER_NORM_DYNAMIC:
+        const float c_max = fabsf(mixer_in->roll) + fabsf(mixer_in->pitch) +
+                            fabsf(mixer_in->yaw);
+
+        const float headroom = 1.0f - mixer_in->thrust;
```
lkijanovic: Shouldn't we be worried about going below zero?
Something like: `headroom = min(thrust, 1-thrust)` might be better.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,234 @@
...
+        }
+    }
+    return MIXER_SUCCESS;
+}
+
+static void normalize_control(MIXER_Inst_Type *mixer,
+                              MIXER_Input_Type *mixer_in) {
+    switch (mixer->norm) {
+    case MIXER_NORM_STATIC:
+        const float scale = 1.0f / 3.0f;
```
DNedic: You use fixed factors here, but motor factors could change, at least you made a function to make them configurable. If we don't want them to be configurable (always perfect 4 motor x configuration) we could just encode it to a const struct.

filipembedded: I changed hardcoded scale here to configurable Thrust to RPY weight ratio in pre-compile time using Kconfig symbol for it. @nikolaptr suggested that Thrust should have higher priority over Pitch and Roll, and Pitch and Roll should have higher priority over Yaw. 
For now, I only implemented T to RPY ratio, but I can adapt it to fully match the @nikolaptr idea.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,185 @@
...
+ *
+ *         FRONT
+ *      [3]     [1]
+ *         \   /
+ *          [ ]
+ *         /   \
+ *      [2]     [4]
+ *         BACK
+ *
+ *      Initialization flow should be done like in this code snippet:
```
DNedic: See previous comment, requiring the user to follow this procedure is just waiting for bugs to happen, and it is useless boilerplate. Bundle the motor info and the configuration info together and pass them to `MIXER_Init()` once

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,236 @@
...
+    return MIXER_SUCCESS;
+}
+
+static void normalize_control(MIXER_Inst_Type *mixer,
+                              MIXER_Input_Type *mixer_in) {
+    switch (mixer->norm) {
+    case MIXER_NORM_STATIC:
+        const float thrust_weight = CONFIG_STATIC_NORM_THRUST_WEIGHT / 100.0f;
+        const float rpy_weight = 1.0f - thrust_weight;
+        mixer_in->thrust *= thrust_weight;
```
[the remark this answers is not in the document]
filipembedded: Good catch!

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,236 @@
...
+        mixer_in->roll *= rpy_weight;
+        mixer_in->pitch *= rpy_weight;
+        mixer_in->yaw *= rpy_weight;
+        break;
+
+    case MIXER_NORM_DYNAMIC:
+        const float c_max = fabsf(mixer_in->roll) + fabsf(mixer_in->pitch) +
```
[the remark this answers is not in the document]
filipembedded: c_max represents the worst-case control magnitude that could affect any single motor.

We use absolute values because each motor has different mixing factors (+-1). For a given motor, the roll/pitch/yaw contributions could all add up in the same direction.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,185 @@
...
+
+typedef struct {
+    float thrust; // 0.0 to 1.0
+    float roll;   // -1.0 to 1.0
+    float pitch;  // -1.0 to 1.0
+    float yaw;    // -1.0 to 1.0
+} MIXER_Input_Type;
+
+typedef struct {
+    float roll_factor;     // -1.0 to 1.0
```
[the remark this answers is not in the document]
filipembedded: Yeah you're right, these comments are useless

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,236 @@
...
+    memset(mixer->motor_factors, 0, sizeof(mixer->motor_factors));
+    memset(mixer->motor_outputs, 0, sizeof(mixer->motor_outputs));
+    mixer->motor_instances = 0;
+    mixer->max_motor_num = 0;
+    mixer->uav_config = uav_cfg;
+    mixer->norm = norm;
+    mixer->initialized = false;
+
+    switch (uav_cfg) {
+    case MIXER_UAV_CFG_QUADROTOR_X:
```
lkijanovic: I somewhat dislike checking for geometry all over the place in this file. Maybe just do a top-level switch/case here and call static functions to init geometry-specific stuff, i.e:
```
switch (uav_cfg) {
  case MIXER_UAV_CFG_QUADROTOR_X:
     setup_quadrotor_x(...);
     break;
  default:
     return MIXER_INVALID_CFG;
}
```

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,236 @@
...
+
+static void normalize_control(MIXER_Inst_Type *mixer,
+                              MIXER_Input_Type *mixer_in);
+
+MIXER_Error_Type MIXER_AddMotor(MIXER_Inst_Type *mixer, ESC_Inst_Type *esc) {
+    if (mixer == NULL || esc == NULL) {
+        return MIXER_INIT_ERROR;
+    }
+
+    if (!mixer->initialized) {
```
lkijanovic: Do we actually need this field? If user didn't call init we're dealing with a garbage value here anyway...

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,236 @@
...
+            input->roll * mixer->motor_factors[i].roll_factor +
+            input->pitch * mixer->motor_factors[i].pitch_factor +
+            input->yaw * mixer->motor_factors[i].yaw_factor;
+    }
+}
+
+static MIXER_Error_Type apply_motor_outputs(MIXER_Inst_Type *mixer) {
+    for (uint8_t i = 0; i < mixer->motor_instances; i++) {
+        ESC_Error_Type esc_status =
+            ESC_SetSpeed(&mixer->motor_arr[i], mixer->motor_outputs[i]);
```
lkijanovic: Hm, I'm not completely sure how I feel about this from an architectural standpoint.
I guess I'm fine with this especially for initial implementation but I'd prefer if mixer didn't "own" ESCs but instead just did input transformation i.e:
`MIXER_Execute(mixer, input, output)`

From my standpoint this has multiple benefits:

- Mixer code is simplified, e.g. `MIXER_AddMotor` API is no longer necessary (@DNedic )
- Mixer can easily be unit-tested completely isolated from the hardware
- Mixer outputs can be logged or emitted via telemetry without polluting either mixer or ESC code

filipembedded: I totally agree. That would make unit tests much easier to implement, and could help with SITL too. I will start implementing this soon

libs/motor_control/mixer/mixer.c
filipembedded: MIXER_AddMotor is ugly API I came up with. @DNedic had some ideas to solve this, where user doesn't need to respect the complicated initialization process. I will probably remove MIXER_AddMotor totally and just have one init call

libs/motor_control/mixer/mixer.c
filipembedded: Valid. I will consider this in a new update

libs/motor_control/mixer/mixer.c
lkijanovic: It's not just about this API, this field is checked in other APIs as well.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * 1. If the torque span (max c_i - min c_i) exceeds the full output range,
+ *    all torques are scaled by a common factor. The torque magnitude is
+ *    reduced, but its direction is preserved exactly.
+ * 2. The collective thrust is clamped into [-min c_i, 1 - max c_i], the
+ *    window in which no motor saturates. Thrust accuracy is sacrificed
+ *    before control authority: attitude control keeps the craft upright,
+ *    while a thrust error only costs altitude.
+ *
+ * Consequence of step 2: with zero commanded thrust and a nonzero torque
```
DNedic: Nitpick: Either I don't fully understand this part or it is just incoherent.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,157 @@
...
+ * demand the motors still spin ("airmode" behavior), and full control
+ * authority is retained even at maximum commanded thrust. Suppressing
+ * outputs while disarmed is the caller's responsibility.
+ *
+ * @retval MIXER_INVALID_ARG An argument is NULL
+ * @retval MIXER_INVALID_CFG Mixer instance holds an unsupported frame
+ *                           configuration (e.g. it was never initialized)
+ * @retval MIXER_SUCCESS Motor outputs computed successfully
+ */
+MIXER_Error_Type MIXER_Execute(const MIXER_Inst_Type *mixer,
```
DNedic: Nitpick: Could we name it "Step" or something of the sort so that it is clear that it just does one iteration and doesn't contain a loop or a thread spawn inside at a glance?

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,157 @@
...
+ * @param input Pointer to the control demands (not modified)
+ * @param output Pointer to the resulting motor outputs
+ *
+ * Each motor output is composed as:
+ *
+ *     m_i = T + c_i
+ *
+ * where T is the collective thrust and c_i is the torque contribution of
+ * motor i (the roll/pitch/yaw demands weighted by the frame's mixing
+ * factors). The torque factors of every axis sum to zero across the motors,
```
DNedic: Nitpick: Torque factors sum to zero only in the case of symmetric airframes, and since we take an airframe type as an argument, we shouldn't say this in the javadoc.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,157 @@
...
+ *           \       /
+ *            \     /
+ *             (   )
+ *            /     \
+ *           /       \
+ *      [1] CW       [3] CCW
+ *              BACK
+ */
+typedef enum {
+    MIXER_QUAD_X_MOTOR_FRONT_RIGHT = 0,
```
DNedic, line 84: For future consideration: What if we encoded these with a coordinate pair instead? We could use that internally straight away without a switch case and on the API side would easily be able to support non-quadx setups.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+MIXER_Error_Type MIXER_Execute(const MIXER_Inst_Type *mixer,
+                               const MIXER_Input_Type *input,
+                               MIXER_Output_Type *output) {
+    if (mixer == NULL || input == NULL || output == NULL) {
+        return MIXER_INVALID_ARG;
+    }
+
+    const MIXER_Frame_Geometry_Type *geometry = frame_geometry(mixer->frame);
```
DNedic: We shouldn't validate this twice, we already validated it in `Init`.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    float torque_min = torque[0];
+    float torque_max = torque[0];
+
+    for (uint8_t i = 1; i < geometry->motor_count; i++) {
+        if (torque[i] < torque_min) {
+            torque_min = torque[i];
+        }
+        if (torque[i] > torque_max) {
```
DNedic: Mega nitpick: If we have a new minimum, it's definitely not a new maximum so we can use `else` here.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /* If the torque span exceeds the full output range, no collective thrust
+       can fit all motors into [0, 1]. Scale all torques by a common factor:
+       the torque magnitude is reduced, but its direction is preserved. */
+    const float span = torque_max - torque_min;
+    if (span > 1.0f) {
+        const float scale = 1.0f / span;
+
+        for (uint8_t i = 0; i < geometry->motor_count; i++) {
```
DNedic: Nitpick: I wouldn't loop on `uint8_t`s since this will cause extra overflow check instructions on any non-8bit platforms causing a performance hit.

libs/motor_control/mixer/mixer.h
filipembedded: I agree, we should be precise

libs/motor_control/mixer/mixer.c
filipembedded: Hmm nice catch. I would change it to `size_t` to cover both 32-bit and 64-bit because SITL will be run on X64 CPU-s

libs/motor_control/mixer/mixer.h
filipembedded: I agree, but I would implement it in a follow-up PR.

libs/motor_control/mixer/mixer.h
filipembedded: I would rename it to the `MIXER_Calculate`, but I don't have an issue with `MIXER_Step` either.

libs/motor_control/mixer/mixer.c
filipembedded: The frame is validated in `MIXER_Init` and is not re-validated here. This call fetches the factor table, which `MIXER_Calculate` cannot work without, and the NULL branch checks a different thing: that the instance was actually initialized. It protects against a caller ignoring the result of `MIXER_Init`, or worse, never calling it at all, and it is what replaced the `initialized` flag we dropped earlier in this review.

DNedic: Makes sense.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,153 @@
...
+
+    /* Adding equal thrust to all motors produces no torque, so the collective
+       thrust can move freely inside the window where no motor leaves [0, 1].
+       Clamping into that window sacrifices thrust accuracy before control
+       authority. */
+    float thrust = input->thrust;
+    if (thrust > 1.0f - torque_max) {
+        thrust = 1.0f - torque_max;
+    }
+    if (thrust < -torque_min) {
```
DNedic: We need an `else` here.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,153 @@
...
+        thrust = -torque_min;
+    }
+
+    for (size_t i = 0; i < geometry->motor_count; i++) {
+        float value = thrust + torque[i];
```
DNedic: Nitpick: `value` is meaningless.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * @brief Per-motor output values, each guaranteed to be within [0.0, 1.0]
+ *
+ * Only the first motor_count entries are valid. The motor array is indexed
+ * by frame position (see MIXER_Quad_X_Motor_Position_Type).
+ */
+typedef struct {
+    float motor[MIXER_MAX_MOTORS];
+    uint8_t motor_count;
```
DNedic: This is weird considering we already set the frame type ourselves. Do we want the output to be self-descriptive or assume the consumer already knows the configuration? In both cases this is wrong, in case 1 we want to pass the frame type and not motor count, in case 2 we don't even need the motor count at all.

libs/motor_control/mixer/tests/mixer_test.c
```diff
@@ -0,0 +1,270 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
+ * Host-side unit tests for the mixer library.
+ *
+ * Build and run:
```
DNedic: Why do we need to teach people how to use CMake here?

libs/motor_control/mixer/tests/mixer_test.c
```diff
@@ -0,0 +1,270 @@
...
+
+#include "mixer.h"
+#include <math.h>
+#include <stdio.h>
+#include <string.h>
+
+static int total_checks = 0;
+static int failed_checks = 0;
+
+#define CHECK(cond)                                                            \
```
DNedic: Could we use catch2 or doctest instead?

libs/motor_control/mixer/tests/mixer_test.c
```diff
@@ -0,0 +1,270 @@
...
+        total_checks++;                                                        \
+        if (!(cond)) {                                                         \
+            failed_checks++;                                                   \
+            printf("FAIL %s:%d: %s\n", __FILE__, __LINE__, #cond);             \
+        }                                                                      \
+    } while (0)
+
+#define CHECK_CLOSE(a, b) CHECK(fabsf((a) - (b)) < 1e-5f)
+
+static MIXER_Output_Type run_mixer(const MIXER_Inst_Type *mixer, float thrust,
```
DNedic: Why do we need such an useless wrapper?

libs/motor_control/mixer/tests/CMakeLists.txt
```diff
@@ -13,20 +13,28 @@
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+# Host-side unit tests for the mixer library. Not part of the Zephyr build:
+#     cmake -S . -B build
```
DNedic: Same.

libs/motor_control/esc/Kconfig
```diff
@@ -27,6 +27,3 @@ config ESC_MAX_PULSE_OFFSET_PERCENT
 
 
 endif # ESC
-
-
-source "Kconfig.zephyr"
```
DNedic, line 32: Why is this removed?

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,153 @@
...
+
+    for (uint8_t i = 1; i < geometry->motor_count; i++) {
+        if (torque[i] < torque_min) {
+            torque_min = torque[i];
+        } else if (torque[i] > torque_max) {
+            torque_max = torque[i];
+        }
+    }
+
+    /* If the torque span exceeds the full output range, no collective thrust
```
DNedic: I don't understand how the DESIRED torque span can exceed already normalized bounds it has to be passed under. It should be an assert, a sanity check, not scale them and have the user think everything is fine.

If the intent was to ensure no clipping happens, clipping can only be detected when collective thrust and torque-derived thrust for each motor is added up, and the correction of the intensity of the combined thrust vector should be done.

libs/motor_control/esc/Kconfig
filipembedded: `Kconfig.zephyr` belongs to the application `Kconfig`, not to a library one. `demos/mixer/Kconfig` sources both `esc/Kconfig` and `Kconfig.zephyr`, so leaving it in the library meant sourcing it twice, and the mixer demo then fails to configure with `RecursionError: maximum recursion depth exceeded`. `app/Kconfig` and `demos/nvs-spi-flash/Kconfig `already do it the right way, the ESC library was the odd one out.

Your question did uncover a real regression though. `demos/esc` never had its own source line and was getting `Kconfig.zephyr` transitively through the library, so after this removal it fails to configure with undefined symbol LOG and friends. Fixed by adding `source "Kconfig.zephyr"` to `demos/esc/Kconfig`, both demos configure now.

libs/motor_control/mixer/mixer.h
filipembedded: You are right here, I refactored this in my latest commit, please check it out

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,99 @@
...
+                                                .yaw = 1.0f},
+            [MIXER_QUAD_X_MOTOR_REAR_LEFT] = {.roll = 1.0f,
+                                              .pitch = 1.0f,
+                                              .yaw = 1.0f},
+            [MIXER_QUAD_X_MOTOR_FRONT_LEFT] = {.roll = 1.0f,
+                                               .pitch = -1.0f,
+                                               .yaw = -1.0f},
+            [MIXER_QUAD_X_MOTOR_REAR_RIGHT] = {.roll = -1.0f,
+                                               .pitch = 1.0f,
+                                               .yaw = -1.0f},
```
DNedic: Important: I get that this is a sample, simplified mathematical model, but yaw factor should definitely not be equal to roll and pitch, or even close, since it is not created by thrust but by aerodynamic effects and requires an order of magnitude more thrust to get the same change.

We can do one of 2 things:
* be able to provide these to the mixer, from the mathematical model, instead of being hardcoded, then provide a best-guess default (10 from my quick research, Gazebo uses such a factor) from the layer up.
* bake the best-guess default in here instead of `1.0` which is horrible

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /* Protection against non/bad-initialized mixer instance */
+    if (mixer->geometry == NULL) {
+        return MIXER_INVALID_CFG;
+    }
+
+    /* Each axis gets an equal share of the headroom; the worst case is all
+       three aligned on the same motor, so the shares must sum to it. */
+    const float rpy_scale = MIXER_TORQUE_HEADROOM / 3.0f;
```
DNedic: Torque headrooms is frame geometry dependent, should be put in there or derived from there.

filipembedded: I agree

DNedic: Sorry, I meant the `3.0f` used here, which affects the scale. The torque headroom is a statically defined split between collective thrust and attitude control thrust.

Also, if we do https://github.com/eurus-project/efc/pull/81/changes#r3625969019 (and we should), the 3.0f becomes 12f or whichever new max we have.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "tools/jMAVSim"]
 	path = tools/jMAVSim
 	url = git@github.com:PX4/jMAVSim.git
+[submodule "doctest"]
+	path = tools/doctest
```
DNedic: `doctest` isn't a tool, it's a library, we should move it.

libs/motor_control/mixer/mixer.c
DNedic: On second thought, that would completely ruin the pitch and roll control since yaw commands ~10 times higher torque to get the same change in angular velocity and hence pulls the statically allocated max for pitch and roll. We do need prioritization, but for a start we can make it a constant factor.

.gitmodules
filipembedded: Moved to libs/third_party

licenserc.toml
```diff
@@ -35,7 +35,8 @@ includes = [
 
 excludes = [
   "tools/mavlink/**",
-  "tools/jMAVSim/**"
+  "tools/jMAVSim/**",
+  "tools/doctest/**"
```
DNedic, line 39: Doctest was moved, move the exclusion too.

libs/motor_control/mixer/mixer.h
```diff
@@ -0,0 +1,209 @@
...
+ * Each motor output is composed as:
+ *
+ *     m_i = T + c_i
+ *
+ * where T is the collective thrust and c_i is the torque contribution of
+ * motor i (the roll/pitch/yaw demands weighted by the frame's mixing
+ * factors). Saturation is prevented statically, by construction, rather than
+ * detected and corrected at run time:
+ *
+ * 1. Each axis demand is scaled by torque_scale, computed once by MIXER_Init
```
DNedic, line 176: Can we not write the whole implementation in the javadoc? It just needs the things the caller needs to worry about and for everything else, the implementation should be self-descriptive as much as possible.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,112 @@
...
+   MIXER_Calculate describes: each motor carries the collective thrust equally
+   (there is no thrust factor, it is implicitly 1.0) and the torque factors of
+   each axis sum to zero across the motors. */
+static const MIXER_Frame_Geometry_Type quad_x_geometry = {
+    .motor_count = 4,
+    .factor =
+        {
+            [MIXER_QUAD_X_MOTOR_FRONT_RIGHT] = {.roll = -1.0f,
+                                                .pitch = -1.0f,
+                                                .yaw = 0.1f},
```
DNedic, line 33: It should be 10, not 0.1. This is how much torque we need to get a certain angular speed, and we know that it takes much more to get yaw than roll and pitch, since yaw is impacted by the torque generated by counter-forces from the motor rotor and propeller rotation.

libs/motor_control/mixer/mixer.c
```diff
@@ -0,0 +1,112 @@
...
+    const float roll = input->roll * mixer->torque_scale;
+    const float pitch = input->pitch * mixer->torque_scale;
+    const float yaw = input->yaw * mixer->torque_scale;
+
+    /* Compresses thrust into [headroom, 1 - headroom] so the torque budget
+       above never pushes a motor outside [0, 1]. */
+    const float thrust = MIXER_TORQUE_HEADROOM +
+                         input->thrust * (1.0f - 2.0f * MIXER_TORQUE_HEADROOM);
+
+    for (size_t i = 0; i < mixer->geometry->motor_count; i++) {
```
DNedic, line 104: Where has the rp prioritization gone?

